PR #5, opened 2017-09-10
Skip eslintignore'd files

### LINKED ISSUE #3: eslintignore'd files should be marked "skip", not "pass"
When files are listed in .eslintignore, jest-runner-eslint marks them as PASS, even when they contain errors.

I'd expect ignored files to either actually *be* ignored, or, for them to show up as skipped tests.

rogeliog:
Fixes #3 

screen shot 2017-09-10 at 9 07 34 am


ljharb:
Thanks! So, to clarify: this means ignored files will not show up at all, but will be counted as skipped?

rogeliog:
Correct, here is a gif of the current proposal, let me know what you think... I'm also happy to modify the behavior is something else makes more sense

eslintignore

ljharb:
I think that works; the only other thing might be actually showing the skipped filenames - since skipped tests still do show up on the output as such.

rogeliog:
It would be interesting to show the skipped file names, but I think that will require a larger refactor and/or changes in `facebook/jest`...

Another option that we could do is: Check if it is ignored by ESLint before calling `onStart`(here)[https://github.com/rogeliog/jest-runner-eslint/blob/master/src/index.js#L41]... Which will and never report back to Jest that these "test files" ever existed. So, basically completely ignoring them


FYI: This is how it looks when I run `jest-runner-eslint` against the `facebook/jest` codebase
screen shot 2017-09-10 at 10 08 55 am

PR #6, opened 2017-09-10
Remove babel-polyfill and drop support for Node 4

### LINKED ISSUE #1: Avoid using babel-polyfill?
babel-polyfill (and thus babel-runtime and core-js) are deps that Airbnb doesn't want anywhere in its graph; is there any way that runtime dep could be avoided?

rogeliog:
Fixes #1 

This PR also drops support for Node 4, we can consider supporting it if it is ever needed.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -5,7 +5,7 @@
       {
         "useBuiltIns": true,
         "targets": {
-          "node": "4"
+          "node": "6"
         }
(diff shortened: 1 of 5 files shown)


ljharb:
node 4 is still LTS, and eslint supports it; it'd be better to retain support for it.

PR #12, opened 2017-09-14
Use array for failureMessages

lencioni:
From what I can tell, this will make the test results more consistent
with jest.

https://github.com/facebook/jest/blob/193d646b/fixtures/failing_jsons/failing_jest_json.json


rogeliog:
Awesome thanks!

Issue #10, opened 2017-09-12
Feature request: provide a way to run with --quiet and --fix options


lencioni:
I sometimes want to run eslint with the `--quiet` and `--fix` options to autofix rules across large numbers of files. I was unable to find a way to use these options with the jest runner and think it would be really great if there was a way.

If there is already a way, it would be wonderful to document it!

rogeliog:
Hi 👋,

Thanks for reporting!  

Yeah this will be real nice. I think this is part of a broader discussion _(How to configure Jest runners?)_, for which there is an ongoing discussion here -> https://github.com/facebook/jest/issues/4278

Off the top of my head I think of some different paths that we could take.
* Fix it in `facebook/jest` and allow for a way to customize runners
* AFAIK passing unknown CLI args to Jest will cause jest to fail with an unrecognized arg error. Maybe there could be a way to pass extra unknown args, but this will get weird when running wiht multiple projects at the same time.
* Come up with a solution similar to [jest-runner-mocha#custom-config-options](https://github.com/rogeliog/jest-runner-mocha#custom-config-options), where we add yet another config file or field in the `package.json`

This doesn't mean that they are the correct solutions, so I'm more than open to better solutions

ljharb:
Can runner options be passed on the command line?

rogeliog:
That would be awesome but right now Jest returns

```
● Unrecognized CLI Parameters:

  Following options were not recognized:
  ["quiet", "fix"]

  CLI Options Documentation:
  http://facebook.github.io/jest/docs/cli.html

error Command failed with exit code 1.
```

ljharb:
That's not what I meant - custom reporters in a jest config can take options https://facebook.github.io/jest/docs/en/configuration.html#reporters-array-modulename-modulename-options - can those be provided via command line at all, in `--reporters`?

rogeliog:
Oh sorry, I misunderstood... my bad!

No, runners can't do that right now, but that is one of the proposals on how to add configs to runners https://github.com/facebook/jest/issues/4278#issuecomment-326676409

PR #13, opened 2017-09-20
Allow config overrides

### LINKED ISSUE #11: Ability to specify lintrc file
Hello,

First off, AWESOME job!  Just implemented this and had a jaw dropping moment of how fast the linting is for our CI.

Is it possible to specify a specific .eslintrc file to run the jest-runner with?

Our scenario is:

 - Legacy linting from jshint converted to eslint
 - CI is ran via this eslintrc file via gulp (.lintrc)
 - Dev IDEs are running off of the root level .eslintrc extending AirBnB's style guide to allow us to slowly convert to the new styleguide we have in place.

I would like to be able to pass in said .lintrc to be able to use the needed linting rules for our CI.  Thoughts?

rogeliog:
This add the ability to pass a `jest-runner-eslint.config.js` file, with full support for https://eslint.org/docs/developer-guide/nodejs-api#cliengine

Fixes #11 and part of #10


README.md
```diff
@@ -44,3 +44,27 @@ module.exports = {
...
+
+This project uses [cosmiconfig](https://github.com/davidtheclark/cosmiconfig), so you can provide config via:
+* a `jest-runner-eslint` property in your `package.json`
+* a `jest-runner-eslint.config.js` JS file
+* a `.jest-runner-eslintrc` JSON file
+
+```js
+module.exports = {
+  "cliOptions": {
+    // ESLint cli options options here
```
ljharb: can you include some examples, especially ones that aren't all lowercase, or have multiple words?

README.md
```diff
@@ -44,3 +44,27 @@ module.exports = {
...
+  },
+  "writeOnFix": // (true|false)
```
ljharb: couldn't this instead be `fix: true` in `cliOptions`?

rogeliog: So my thought process was this: The ESLint Node api has a [`fix` option](https://eslint.org/docs/developer-guide/nodejs-api#cliengine) but it works a bit different than `--fix` because it doesn't save to disk the result... In order to do that `outputFixes()` needs to be called.

One options that I thought is: "Always call `outputFixes()` when `fix === true`"... but I was a bit worried about it because then it would break the 1:1 mapping with the ESLint docs... If you think that is a better option I'm happy to change it 😄

rogeliog: Also, I agree that the `writeOnFix` feels a awkward so I'm happy to change it

ljharb: Since this is eslint command line options, and on the CLI, `--fix` definitely writes to disk, I think that cliOptions is the right place to put it.

In other words, `cliOptions` to me maps to what `eslint` on the CLI supports - what the eslint API supports won't actually matter to users.

rogeliog: Perfect, I agree!

README.md
```diff
@@ -44,3 +44,60 @@ module.exports = {
...
+  }
+}
+```
+
+#### `globals`
+This is similar to running ESLint with `--globals`
+```json
+{
+  "jest-runner-eslint": {
+    "cliOptions": {
```
rogeliog: @ljharb I wonder if it is even worth having it nested under `cliOptions`?

I originally did it so that it is a bit more feature proof in case we add options that do no map to `cliOptions`, but it might be an overkill

ljharb: Having been burned by this in the past, I think making `cliOptions` is definitely prudent, even if it's a bit annoying at times for users.

I'd much rather have an unneeded `cliOptions`, than need to add a new kind of option and have no place to put it :-)

README.md
```diff
@@ -44,3 +44,60 @@ module.exports = {
...
+In `package.json`
+```json
+{
+  "jest-runner-eslint": {
+    "cliOptions": {}
+  }
+}
+```
+
+See https://eslint.org/docs/developer-guide/nodejs-api#cliengine for a full
```
ljharb: I think `cliOptions` should actually map to https://eslint.org/docs/user-guide/command-line-interface

README.md
```diff
@@ -44,3 +44,60 @@ module.exports = {
...
+}
+```
+
+#### `configFile`
+This is similar to running ESLint with `-c /path/to/config-file`
+```json
+{
+  "jest-runner-eslint": {
+    "cliOptions": {
+      "configFile": "/path/to/config-file"
```
ljharb: similarly, this should be `config`, since it should map to actual CLI options, and not to eslint's API (that very few users ever interact with)

README.md
rogeliog: I think that is a good point.

Even though it would be more work, because we need to make sure that we do the correct mapping of "pseudo CLI options" to API options, but it will definitely provide a better experience.

After looking at the docs here is the list of CLI options that we could have. It does not include all of them because the API does not accept all of them. Although a lower priority, but we might need to shim the rest

|Proposed option | ESLint CLI option| Notes|
|:---------------|:-----|:-----|
|noInlineConfig | --no-inline-noInlineConfig|
|cache | --cache|
|cacheLocation | --cache-location|
|config | --config|
|envs | --env | _Needs to be pluralized because `--env` can be passed multiple times_
|exts | --ext | _Needs to be pluralized because `--ext` can be passed multiple times_
|fix | --fix|
|globals | --global | _Needs to be pluralized because `--global` can be passed multiple times_
|noIgnore | --no-ingore|
|ignorePath | --ignore-ignorePath|
|parser | --parser|
|parserOptions | --parser-parserOptions|
|plugins | --plugin | _Needs to be pluralized because `--plugin` can be passed multiple times_
|rulesdirs | --rulesdir | _Needs to be pluralized because `--rulesdir` can be passed multiple times_
|rules | --rule | _Needs to be pluralized because `--rule` can be passed multiple times_
|noEslintrc | --no-eslintrc|
|exts | --ext  | _Needs to be pluralized because `--exts` can be passed multiple times_


What do you think?

ljharb: I don't think they need to be pluralized; I think if they want more than one, it can be an array - like `ext: 'foo'` or `ext: ['foo', 'bar']`

PR #14, opened 2017-11-06
Use caret constraint for `eslint` dependency

Turbo87:
For some unknown reason `eslint` was pinned to v4.5.0, but since the tests are running fine with the latest release we should lift that constraint.


Turbo87:
@rogeliog this is rebased and 🍏  now too :)

PR #24, opened 2017-12-28
Remove parser and parserOptions from normalizeConfig

### LINKED ISSUE #19: Parser option ignored in eslint config.
Hi,

I converted our project to use the jest eslint test runner instead of eslint directly but have a weird issue. 

Using eslint with `parser` set to `babel-eslint` in the config file worked perfectly, but when using the jest-runner, It doesn't seems to pick it up (I get unexpected token issues, but the rest of the config works).

If I override the `parser` in `cliOptions` then it works but it should take the one from my `.eslintrc.js` in the first place.

Thanks !

azz:
Not sure if this is the correct fix (potentially more options can be removed from here for the same reason). Thought I should open a PR to fix #19.


azz:
Done. Rather than adding a test for another parser (like `babel-eslint`) to test a language feature that may change in the future I added a test for a parser that doesn't exist to confirm that it's not overwritten with `espree`.

azz:
Good to merge?

rogeliog:
@azz I just released a pre-minor version `@0.4.0-0` also available `@next` let me know how that is working for you and we ca release `0.4.0` after that.

It includes #24 and #25

fdubost:
Hi ! It works for me 👍

PR #32, opened 2018-02-03
Use `jest-worker`.

### LINKED ISSUE #23: Switch from `worker-farm` to `jest-worker`?
In Jest 22 the team exposed the internal `worker-farm` equivalent, [`jest-worker`](https://github.com/facebook/jest/tree/master/packages/jest-worker). Would it be worth switching from `worker-farm` to `jest-worker` for consistency with Jest's default runner?

wtgtybhertgeghgtwtg:
Maybe closes #23.  I'm not too confident about it, though.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,6 +1,5 @@
 const throat = require('throat');
-const pify = require('pify');
-const workerFarm = require('worker-farm');
+const Worker = require('jest-worker').default;
 const path = require('path');
 
 const TEST_WORKER_PATH = path.join(__dirname, 'runESLint.js');
@@ -19,18 +18,12 @@ module.exports = class ESLintTestRunner {
 
   // eslint-disable-next-line
   async runTests(tests, watcher, onStart, onResult, onFailure) {
-    const farm = workerFarm(
-      {
-        autoStart: true,
-        maxConcurrentCallsPerWorker: 1,
-        maxConcurrentWorkers: this._globalConfig.maxWorkers,
-        maxRetries: 2, // Allow for a couple of transient errors.
-      },
-      TEST_WORKER_PATH,
-    );
+    const worker = new Worker(require.resolve(TEST_WORKER_PATH), {
+      maxRetries: 2,
+      maxWorkers: this._globalConfig.maxWorkers,
+    });
 
     const mutex = throat(this._globalConfig.maxWorkers);
-    const worker = pify(farm);
 
     const runTestInWorker = test =>
       mutex(async () => {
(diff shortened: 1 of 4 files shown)


src/index.js
```diff
@@ -1,6 +1,5 @@
 const throat = require('throat');
-const pify = require('pify');
-const workerFarm = require('worker-farm');
+const Worker = require('jest-worker').default;
```
ljharb: `.default` is a babel implementation detail, and should never be exposed in an entry point. is there a bug filed on `jest-worker` for this?

src/runESLint.js
```diff
@@ -59,43 +59,32 @@ const pass = ({ start, end, testPath }) =>
     jestTestPath: testPath,
   });
 
-const runESLint = ({ testPath, config }, workerCallback) => {
-  try {
-    const start = +new Date();
+module.exports.runESLint = ({ testPath, config }) => {
+  const start = +new Date();
```
ljharb: `Date.now()`

src/runESLint.js
```diff
@@ -59,43 +59,32 @@ const pass = ({ start, end, testPath }) =>
...
+  }
 
-      if (options.cliOptions && options.cliOptions.fix) {
-        CLIEngine.outputFixes(report);
-      }
+  const { CLIEngine } = getLocalESLint(config);
+  const options = getESLintOptions(config);
+  const cli = new CLIEngine(options.cliOptions);
+  if (cli.isPathIgnored(testPath)) {
+    const end = +new Date();
```
ljharb: also `Date.now()`

src/runESLint.js
```diff
@@ -59,43 +59,32 @@ const pass = ({ start, end, testPath }) =>
...
+  if (options.cliOptions && options.cliOptions.fix) {
+    CLIEngine.outputFixes(report);
+  }
 
-      if (report.errorCount > 0) {
-        const formatter = cli.getFormatter(options.cliOptions.format);
-        const errorMessage = formatter(
-          CLIEngine.getErrorResults(report.results),
-        );
+  const end = +new Date();
```
ljharb: also here

src/index.js
wtgtybhertgeghgtwtg: I don't see any.

src/runESLint.js
wtgtybhertgeghgtwtg: It was like that when I got here.  Should I still fix it?

src/index.js
ljharb: Would you mind filing one and linking to it here?

src/runESLint.js
ljharb: Ah, fair enough. Not required, but it’d be nice :-)

src/index.js
wtgtybhertgeghgtwtg: I'm not very good with words, but I don't mind.  What should I say in this issue?

midzelis: Whats up with this patch. It seems good to me. I don't see an issue with the .default thing. In fact, its how https://github.com/rogeliog/create-jest-runner/blob/master/lib/createJestRunner.js#L1 does it, which was recommended in issue #23

ljharb: I suppose it doesn't have to block this issue; but it's indeed a huge problem for babel interop details to pollute the ecosystem.

I've filed https://github.com/facebook/jest/issues/5803

PR #34, opened 2018-02-16
Document usage with other Jest runners

### LINKED ISSUE #22: Document how to set this up alongside your regular jest config
1. make two jest configs
  - one with your regular configuration
  - another one with eslint configuration
2. add `projects: ['path/to/jest-main.config.js', 'path/to/jest-eslint.config.js']` in the regular config file for Jest

Is that the recommended way?

macklinu:
Resolves #22

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,8 @@ npm install --save-dev jest jest-runner-eslint
 
 ### Add it to your Jest config
 
+#### Standalone
+
 In your `package.json`
 ```json
 {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -54,6 +56,88 @@ module.exports = {
 
 Please update `testMatch` to match your project folder structure
 
+#### Alongside other runners
```
rogeliog: This is awesome! I think as a next step it would be awesome to make a PR to `facebook/jest` and copy this as the documentation for Jest Projects. Once that happens we can just link to those docs from here

azz:
@rogeliog  I already added some docs on the jest side here: https://facebook.github.io/jest/docs/en/configuration.html#projects-array-string-projectconfig, but it could always be improved.

PR #35, opened 2018-03-01
Use env var to override local eslint location

midzelis:
The rationale behind this is we have a non-standard project set up. We have a lot of .js files that are not part of a npm project, and hence don't have any /node_modules/ anywhere in the ancestor path. 

We work around this for jest transformers/resolvers by always specifying the full filesystem path to the node module, instead of just relying on the node resolution algorithm to discover modules. 

When trying to use this eslint-runner, the eslint location can not be found. This is a simple change that lets one to use the env var ESLINT_PATH to override the location of the local eslint. Had do use the env, since jest doesn't pass any options to the test runners (unfortunately).

### DIFF
--- a/src/utils/getLocalESLint.js
+++ b/src/utils/getLocalESLint.js
@@ -2,6 +2,10 @@ const path = require('path');
 const findUp = require('find-up');
 
 const getLocalESLint = config => {
+  if (process.env.ESLINT_PATH) {
+    // eslint-disable-next-line import/no-dynamic-require,global-require
+    return require(process.env.ESLINT_PATH);
+  }
   const nodeModulesPath = findUp.sync('node_modules', { cwd: config.rootDir });
   // eslint-disable-next-line import/no-dynamic-require, global-require
   return require(path.resolve(nodeModulesPath, 'eslint'));


midzelis:
New changes - instead of using env var, just fallback to normal require('eslint'). This should be a very appropriate fallback.

ljharb:
While this is better, I don't understand how this could ever work without an `eslint` dir in a `node_modules` directory.

midzelis:
It will work if you specify ESLint in the NODE_PATH. See: https://nodejs.org/api/modules.html#modules_loading_from_the_global_folders and https://nodejs.org/api/modules.html#modules_all_together

Note: jest-runner-eslint currently specifies ESLint as a dependency, so it will load from here, before going to NODE_PATH. This is fine with me, since I don't have any version requirements on ESLint, and any version is ok. But it might make sense to move ESLint from dependencies to devDependencies too.

ljharb:
`NODE_PATH` is deprecated though, and won't be supported with ESM use cases - I also don't think we should be making changes to encourage usage of that feature, and ideally, not supporting it at all.

src/utils/getLocalESLint.js
```diff
@@ -3,8 +3,19 @@ const findUp = require('find-up');
 
 const getLocalESLint = config => {
   const nodeModulesPath = findUp.sync('node_modules', { cwd: config.rootDir });
-  // eslint-disable-next-line import/no-dynamic-require, global-require
-  return require(path.resolve(nodeModulesPath, 'eslint'));
+  try {
+    // eslint-disable-next-line import/no-dynamic-require, global-require
+    return require(path.resolve(nodeModulesPath, 'eslint'));
```
SimenB: what about just using https://www.npmjs.com/package/resolve-from here?

Replace the whole `getLocalESLint` with just a `resolveFrom(config.rootDir, 'eslint');`.

I think that should be enough (and it means `findUp` should not be needed)

ljharb: I’d prefer not to use that package due to its aggressive dropping of support of older platforms. Could `resolve` work here?

SimenB: seems like it takes `basedir`, so `resolve.sync('eslint', { basedir: config.rootDir })` should in theory do the trick as well

PR #36, opened 2018-03-15
Implement support for —runInBand

### LINKED ISSUE #20: Support Jest's `runInBand` option
Was looking at trying to debug #19, but realised that the runner always uses a worker-farm and ignores Jest's `runInBand` option - this makes it hard to debug the runner.

Might look at adding this option first :)

midzelis:
Fixes #20

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -2,6 +2,7 @@ const throat = require('throat');
 const pify = require('pify');
 const workerFarm = require('worker-farm');
 const path = require('path');
+const lint = require('./runESLint');
 
 const TEST_WORKER_PATH = path.join(__dirname, 'runESLint.js');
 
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -12,13 +13,41 @@ class CancelRun extends Error {
...
+            } else {
+              resolve(result);
+            }
+          },
+        );
+      });
+    });
+  };
+  return Promise.all(
+    tests.map(test =>
```
ljharb: Do you want these to be serial, or single-threaded?

This approach is single-threaded, but is also concurrent. If you wanted serial execution, you'd need to reduce over `tests` instead of using Promise.all + map.

The function name should match the implementation :-)

rogeliog:
I think we can start using https://github.com/rogeliog/create-jest-runner here. @SimenB just merged a PR for adding runInBand to it https://github.com/rogeliog/create-jest-runner/pull/6

midzelis:
works for me

SimenB:
See #37

Issue #27, opened 2018-01-18
Unknown option "cliOptions" + cache


fdubost:
Hi !

When I use the `cliOptions` params, I get this error :

```
Unknown option "cliOptions"  with value {"cacheLocation": ".eslintcache"} was found.
```

jest@22.1.2
jest-runner-eslint@0.4.0-0

On the other hand, cache doesn't seems to work :
- no `.eslintcache` file is created,
- when I run the command a second time, it take the same time.

Thanks for your help !

ljharb:
What’s your full jest config?

fdubost:
An example here :
```js
const path = require('path');
const rootDir = path.join(__dirname, '../../');

module.exports = {
  runner: 'jest-runner-eslint',
  rootDir: `${rootDir}`,
  displayName: 'lint',
  testMatch: ['<rootDir>/src/**/*.js'],
  cliOptions: {
    cacheLocation: '.eslintcache',
  },
};

```

rogeliog:
Try putting this in your package.json

```json
"jest-runner-eslint": {
  "cliOptions": {
    "cacheLocation": ".eslintcache"
  }
}
```

fdubost:
@rogeliog I have no error but no `.eslintcache` file is created... 😥

dferber90:
It's not pointed out clearly in the docs that there are two configuration files.
If I read them correctly, it seems like you're mixing the config files up.

The two config files:
- `jest.config.js`
- `jest-runner-eslint.config.js`

The second one would take the `cliOptions` part.

However I couldn't get it to work with custom ESLint rules either :/

tdeekens:
This is the current setup in a `jest-runner-eslint.config.js` file.

```
module.exports = {
  cliOptions: {
    format: 'node_modules/eslint-formatter-pretty',
    rule: [
      'import/no-unresolved: 2',
      'prettier/prettier: ["error", {"trailingComma": "es5", "singleQuote": true}]',
    ],
  },
};
```

The config seems to end up in ESLint correctly and is parsed nicely. 

```
{ cacheLocation: '.eslintcache',
     configFile: null,
     envs: [],
     extensions: [ '.js' ],
     fix: false,
     format: 'node_modules/eslint-formatter-pretty',
     globals: [],
     ignorePath: null,
     useEslintrc: true,
     ignore: true,
     allowInlineConfig: true,
     parser: null,
     parserOptions: null,
     plugins: [],
     rules:
[...]

PR #39, opened 2018-04-04
fix(cli-options): rules should be passed as an object instead of array of strings

emmenko:
Fixes: https://github.com/jest-community/jest-runner-eslint/issues/27#issuecomment-375263725

In the [ESLint docs for the CLIEngine](https://eslint.org/docs/developer-guide/nodejs-api#cliengine) clearly states that `rules` is a JSON object and not an array of strings.

The "string" version (e.g. `--rule "quotes: [2, double]"`) is only used by the CLI. When the args are parsed, the `rule` is also a JSON object.

In the case of this library, the `jest-runner-eslint.config.js` is a js file, so we can pass an object directly.

```js
module.exports = {
  cliOptions: {
    format: 'node_modules/eslint-formatter-pretty',
    rules: {
      'import/no-unresolved': 2,
      'prettier/prettier': [
        'error',
        { trailingComma: 'es5', singleQuote: true },
      ],
    },
  },
};
```

Now ESLint properly receives the additional rules.

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,5 +193,5 @@ jest-runner-eslint maps a lot of ESLint CLI arguments to config options. For exa
 |parser|`espree`|`"parser": "flow"`
 |parserOptions|`{}`|`"parserOptions": { "myOption": true }`
 |plugin|`[]`|`"plugin": "prettier"` or `"plugin": ["pettier", "other"]`
-|rule|`null`|`"rule": "'quotes: [2, double]'"` or `"rule": ["quotes: [2, double]", "no-console: 2"]`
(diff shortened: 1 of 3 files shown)


tdeekens:
Thanks a lot for the fix. Any update or release plan on this? Would love to use this fix on a side project of mine too.

emmenko:
@rogeliog any chance we can get this merged? 😇

emmenko:
Awesome, thanks a lot! 🙌

rogeliog:
Awesome, just published `0.6.0` 😄

tdeekens:
Horray. Thanks all. #teamwork

PR #42, opened 2018-04-25
fix: eslint module path resolver for yarn workspaces

### LINKED ISSUE #41: Yarn workspaces
The module resolver looking for your local eslint installation seems not to be compatible with yarn workspaces.

It apparently only looks into the ./node_modules folder in the rootDir and not the repo-root node_modules.

igl:
Instead of looking for the closest node_modules, it will resolve the path like require.resolve() starting from config.rootDir.

Fixes #41
All tests pass
Rejoice

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,8 +24,8 @@
   "dependencies": {
     "cosmiconfig": "^3.0.1",
     "eslint": "^4.5.0",
-    "find-up": "^2.1.0",
     "pify": "3.0.0",
+    "resolve": "1.7.1",
     "throat": "4.1.0",
     "worker-farm": "1.5.0"
   },
--- a/src/utils/__tests__/getLocalESLint.test.js
+++ b/src/utils/__tests__/getLocalESLint.test.js
@@ -1,0 +1,12 @@
+const getLocalESLint = require('../getLocalESLint');
+const requiredESLint = require('eslint');
+
+it('requires eslint', () => {
+  // eslint-disable-next-line global-require
+  expect(getLocalESLint({})).toBe(requiredESLint);
+});
+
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -24,8 +24,8 @@
   "dependencies": {
     "cosmiconfig": "^3.0.1",
     "eslint": "^4.5.0",
-    "find-up": "^2.1.0",
     "pify": "3.0.0",
+    "resolve-from": "4.0.0",
```
ljharb: Please do not se this module; you should be able to achieve this with `resolve`

ljharb: I maintain it; I didn't create it.

No, it doesn't fall back to the built in function, but that's an implementation detail.

igl: tomato potato. Jest actually comes with it's own `jest-resolve` package which seems to be a fork of substack/node-resolve.

They did not write docs about it at all, so i would refrain from using that anyhow.

igl:
It's using `resolve` now.
I added a basic unit test.

Should i do the work and add a proper integration test to check my mono-repo scenario? 
I will have to allow a fake node_modules folder to be in the fixtures folder.
Not sure how to go about that yet. 
Maybe programmatically create it when the test runs?

ljharb:
I think it’s fine to add a fake node_modules and package.json into the repo inside the tests dir

tomasz-sodzawiczny:
@igl Why is this PR stuck? Can I help somehow? I'm also waiting for this to work with yarn workspaces. :)

ljharb:
@igl are you no longer interested in pursuing this?

sibelius:
can't eslint solve this?

Issue #43, opened 2018-05-31
Watch Mode Plugin: jest-runner-eslint/watch


azz:
Combining https://github.com/facebook/jest/issues/5212 and the new Jest Watch Plugins API, it would be really cool to have:

```json
{
  "jest": {
    "runner": "jest-runner-eslint",
    "watchPlugins": [
      "jest-runner-eslint/watch"
    ]
  }
}
```

Which, when an ESLint error is reported, exposes an option in watch mode:

```
Press F to run eslint --fix
```

I think it makes sense for it to live in this repo, but it could equally be a standalone repo.

I'd be happy to work on a PR if it is desired here 😄 

Related: #10

ljharb:
I think it makes sense here.

rogeliog:
Yes, this sounds really interesting! Currently, I don't think that there is a good way of sending that information from the plugin to the runner. It would be awesome find a way that could enable this type of features!

rogeliog:
I think I have some vague idea, I'll try to hack something today

rogeliog:
I think having it in the same repo would make a lot of sense, given that it is a watch plugin really specific to this runner.


#### Here is a PoC:
There are still a couple of hiccups with the implementation, but I first wanted to be on the same page from a product perspective.

It is still pretty rough but I would love to get your thoughts 🙃@ljharb, @azz 

- I added single entrypoint for the watch plugin and then a list of options that could be "modified". I hope that this would allow the plugin to scale better. The UI is similar to https://github.com/rogeliog/jest-watch-select-projects.
- I was wondering what would be other interesting options, that we would like to add to this?
    - `--fix` is implemented and working
    - `--quiet` is still mocked, and does not do anything
- Given that they both live in the same repo, we can share state way easier, which is how the plugin and runner are communicating.
- In order to achieve this there are some small changes that are needed in `create-jest-runner`(to pass extra options to the runner) and there are no changes required in`facebook/jest`.

![watch-eslint](https://user-images.githubusercontent.com/574806/40853446-efac1482-6582-11e8-929f-0f2af23610bc.gif)

@SimenB any thoughts on this?

SimenB:
This is insanely awesome! I think it'd be cool to attach it to `-u`, but maybe not?

>there are no changes required in`facebook/jest`

😍

azz:
Nice! I can't think of any other ESLint flags that I would want to change during a run.

ljharb:
What about `ext`?

[...]

PR #48, opened 2018-09-23
Upgrade Jest and drop Node 4

rogeliog:


SimenB:
@ljharb yes, which is needed for #43

rogeliog:
@ljharb, it is just a `devDepencency` so it is just the version of Jest used to test this runner.

rogeliog:
I guess it is fine to drop Node 4 now that its end of life has come https://github.com/nodejs/Release cc: @ljharb

.travis.yml
```diff
@@ -1,6 +1,5 @@
 language: node_js
 node_js:
-  - "8"
-  - "7"
-  - "6"
-  - "4"
+  - '8'
+  - '7'
```
SimenB: 7 is EOL as well.

Should add 10

ljharb:
A node version being EOL means you *could* drop it; in no way does it mean one *should*.

However, jest’s aggressive dropping of node versions might force our hand, as a jest plugin.

rogeliog:
@ljharb I agree 😄 ! As you mentioned, I meant this in the context of Jest already not supporting Node 4 https://github.com/facebook/jest/pull/4769

PR #53, opened 2018-09-28
Watch plugin for toggling --fix

### LINKED ISSUE #43: Watch Mode Plugin: jest-runner-eslint/watch

rogeliog:
Fixes #43 

I was originally thinking of a more fully featured watch plugin https://github.com/jest-community/jest-runner-eslint/issues/43#issuecomment-395504661, but I was not happy with that UX.

Given that `--fix` is the most common request, I think it makes sense to start with a `--fix` plugin, and if need be we can work on another one later.

fix

- This PR depends on #49, so disregard the babel@7 stuff, I'll rebase once that is merged.
- I intentionally left the plugin in `jest-runner-eslint/watch-fix` instead of `jest-runner-eslint/watch` to leave room for more plugins.
- The way the plugin and the runner communicate is through `configOverrides`, I would love to get some feedback on this. I'm not a fan of it, but we need a way that they can both access shared state... Other ideas are more than welcome.
- One limitation or known issue is: We don't know initially what the current value of `--fix` is in the config. Given this, I decided to show a different message until the user enables the toggling. I think that the UX is still fine.


src/watchFixPlugin/index.js
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  getUsageInfo() {
+    return {
+      key: this._key,
+      prompt: this._getPrompt(),
+    };
+  }
+
+  async _showLoading(fix) {
```
SimenB: why do we need this?

src/runner/__tests__/runESLint.test.js
```diff
@@ -24,7 +24,7 @@ const runESLintRunnerWithMockedEngine = options => {
   });
   const runESLint = require('../runESLint');
 
-  return runESLint(options.runESLint);
+  return runESLint(Object.assign({ extraOptions: {} }, options.runESLint));
```
SimenB: object spread?

README.md
```diff
@@ -172,6 +193,8 @@ module.exports = {
 }
 ```
 
+###
```
SimenB: ?

src/runner/runESLint.js
```diff
@@ -18,12 +18,18 @@ const runESLint = ({ testPath, config }) => {
   }
 
   const { CLIEngine } = getLocalESLint(config);
-  const { cliOptions } = getESLintOptions(config);
-  const cli = new CLIEngine(
-    Object.assign({}, cliOptions, {
-      fix: getComputedFixValue(cliOptions),
-    }),
+  const { cliOptions: baseCliOptions } = getESLintOptions(config);
+  const cliOptions = Object.assign(
```
SimenB: object spread

src/watchFixPlugin/__tests__/index.test.js
```diff
@@ -0,0 +1,73 @@
...
+jest.doMock('chalk', () => new chalk.constructor({ enabled: false }));
+
+jest.useFakeTimers();
+
+let WatchFixPlugin;
+let configOverrides;
+
+const runAndAwait = async plugin => {
+  const runPromise = plugin.run();
+  await jest.runTimersToTime(700);
```
SimenB: we have `runAllPendingTimers` and `runAllTimers` instead of hardcoding the values

SimenB: They also don't return promises, so you can drop the `await`

src/watchFixPlugin/index.js
```diff
@@ -0,0 +1,52 @@
...
+
+const wait = time => new Promise(res => setTimeout(res, time));
+
+class ESLintWatchFixPlugin {
+  constructor({ stdout, config }) {
+    this._stdout = stdout;
+    this._key = config.key || 'F';
+  }
+
+  // eslint-disable-next-line class-methods-use-this
```
SimenB: needed? It _does_ use `this`

src/watchFixPlugin/index.js
SimenB: It looks slick, but does it do anything? Other than delay for a tiny bit

README.md
rogeliog: 🤷‍♂️

src/watchFixPlugin/index.js
rogeliog: The only reason was to provide an indicator and feedback of what is happening, but I guess you are right... none of the other Jest keys does that, I just removed it.

rogeliog:
@ljharb Any thoughts?

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "jest-runner-eslint",
   "version": "0.6.0",
-  "main": "build/index.js",
+  "main": "build/runner/index.js",
```
ljharb: this can just be build/runner

src/watchFixPlugin/index.js
```diff
@@ -0,0 +1,37 @@
...
+
+  getUsageInfo() {
+    return {
+      key: this._key,
+      prompt: this._getPrompt(),
+    };
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  _getPrompt() {
```
ljharb: why can’t this be a closed over function instead of a fake private method?

rogeliog: 👍

src/watchFixPlugin/index.js
```diff
@@ -15,23 +15,22 @@ class ESLintWatchFixPlugin {
   }
 
   getUsageInfo() {
+    const getPrompt = () => {
```
ljharb: if this is defined at module level, then it should be a bit faster.

SimenB:
(we should do #54 before releasing this, BTW)

rogeliog:
@SimenB Perfect!

PR #55, opened 2018-10-02
Move Jest and ESLint to peerDependencies

### LINKED ISSUE #54: Set eslint as a peer dependency
This came up in #49. Right now we depend on ESLint. Not sure exactly why TBH, but my guess is that so we have something to fall back to if we cannot find the user's version.

I think it makes more sense to add it as a peer dependency (with range on both eslint 4 and 5 probably), then we can just use `require` and not have to deal with it manually.

We can also add a peer dep range to Jest as well, especially seeing as #53 requires Jest 23

EDIT: Should resolve #41 I think

rogeliog:
Fixes #54 

I kept the Jest version at 21 given that the runner works well with Jest 21 and Jest 23 is only needed for the watch plugin. I added the appropriate error for when the user tries to use the watch plugin with an insufficient version of Jest.


package.json
```diff
@@ -47,6 +47,10 @@
     "prettier": "^1.14.3",
     "rimraf": "^2.6.2"
   },
+  "peerDependencies": {
+    "eslint": "^4.0.0 || ^5.0.0",
```
SimenB: can you remove `eslint` from dependencies with this change? Also, remove https://github.com/jest-community/jest-runner-eslint/blob/master/src/utils/getLocalESLint.js?

rogeliog: Right, I just moved it to `devDeps`...

I guess https://github.com/jest-community/jest-runner-eslint/blob/master/src/utils/getLocalESLint.js now just becomes a regular `require('eslint')` right?

SimenB: yeah, that's the theory!

package.json
```diff
@@ -26,7 +26,6 @@
     "chalk": "^2.4.1",
     "cosmiconfig": "^5.0.0",
     "create-jest-runner": "^0.4.1",
-    "eslint": "^5.6.0",
     "find-up": "^3.0.0"
```
SimenB: not in use anymore

SimenB: no, I mean `find-up` 🙂 _That's_ not used anymore

package.json
ljharb: It can be both a dep and a peer dep, or a peer and a dev - in either case it will do what you expect. The difference, though, is that if it’s a dep + peer, then it might not be a breaking change to add the peer - if it’s a peer+dev, then it definitely is breaking.

package.json
```diff
@@ -47,6 +47,10 @@
     "prettier": "^1.14.3",
     "rimraf": "^2.6.2"
   },
+  "peerDependencies": {
+    "eslint": "^4.0.0 || ^5.0.0",
+    "jest": ">=21.0.0"
```
ljharb: this should be `^21 || ^22 || ^23` since we cant know if it works with v24 yet.

package.json
rogeliog: @ljharb thanks, you are great!

SimenB: when it's in `dependencies` now, we will always require our own, and never the user's. Which means that this now moves from eslint 4 to eslint 5, which is a breaking change either way.

I suppose we could roll back to eslint v4, but we really should allow people to use eslint v5 with this plugin

ljharb: Not when it's also in peer dependencies - if the user specifies eslint 4, then we won't install v5, we'll prefer their v4.

PR #62, opened 2019-01-26
Update peerDependency jest version range

unindented:
@SimenB I haven't seen any issues after upgrading to `jest@24.0.0`, so this looks like a safe change?


SimenB:
We haven't purposefully changed anything with runners, so should work just fine

ethernal:
When can we expect a release? This one prevents me from updating to Jest 24 and fixing  ``--no-cache --runInBand` error.

SimenB:
`jest-runner-eslint 0.7.3 published 🎉`

PR #65, opened 2019-06-08
chore: bump dated deps

### LINKED ISSUE #64: Reports "NaN total"
Heya!

I just notice that

```
Test Suites: 32 passed, 32 total
Tests:       32 passed, NaN total
Snapshots:   0 total
Time:        3.635s
```

I'm not sure if it was a bug before upgrading to Jest 24.

I may look later on the source code.

SimenB:
GitHub complains about alot of the dated deps in the lockfiles. This hopefully helps 🙂 

Fixes #64


package.json
```diff
@@ -26,24 +26,23 @@
...
-    "eslint-config-prettier": "^3.1.0",
+    "eslint-config-prettier": "^4.3.0",
     "eslint-plugin-import": "^2.14.0",
-    "eslint-plugin-jest": "^21.22.1",
-    "eslint-plugin-prettier": "^2.6.2",
+    "eslint-plugin-jest": "^22.6.4",
+    "eslint-plugin-prettier": "^3.1.0",
     "execa": "^1.0.0",
-    "jest": "^23.6.0",
+    "jest": "^24.8.0",
```
ljharb: This should match the peer dep range verbatim 
```suggestion
    "jest": "^21.0.0 || ^22.0.0 || ^23.0.0 || ^24.0.0",
```
so that `npm ls` passes when testing older jest versions

SimenB:
@ljharb yeah, could probably do that... I like it for the quick installs, but that doesn't really matter on such a small module such as this

ljharb:
it doesn’t even save all that much time in my experience, even on large repos

SimenB:
It used to, might have changed with the newer versions of npm and yarn - I haven't _not_ used a lockfile in a few years 😅
It is quicker for monorepos at least, which is both what both my day job and jest uses, which are the two projects I work on the most. But I'm certainly not against removing them from this repo.

@rogeliog thoughts on this?

ljharb:
ftr I always use a lockfile in a top-level app; just never in anything published, monorepo or not.

PR #70, opened 2019-08-07
expand peerDep to allow eslint@6

kentcdodds:
I'm using this with ESLint 6 in my project and it works fine 👍

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "chalk": "^2.4.1",
     "cosmiconfig": "^5.0.0",
     "create-jest-runner": "^0.5.3",
-    "eslint": "^5.6.0"
+    "eslint": "^4.0.0 || ^5.0.0 || ^6.0.0"
   },
   "devDependencies": {
     "@babel/cli": "^7.1.0",
@@ -47,7 +47,7 @@
     "rimraf": "^2.6.2"
   },
   "peerDependencies": {
-    "eslint": "^4.0.0 || ^5.0.0",
+    "eslint": "^4.0.0 || ^5.0.0 || ^6.0.0",
     "jest": "^21.0.0 || ^22.0.0 || ^23.0.0 || ^24.0.0"
   },
   "prettier": {
--- a/src/utils/normalizeConfig.js
+++ b/src/utils/normalizeConfig.js
@@ -77,7 +77,7 @@ const BASE_CONFIG = {
     default: null,
   },
   parserOptions: {
-    default: null,
+    default: {},
   },
   plugin: {
     name: 'plugins',


kentcdodds:
Actually, because it has a `peerDep` it shouldn't have eslint listed as a `dep` right?

package.json
```diff
@@ -26,14 +26,14 @@
   "dependencies": {
     "chalk": "^2.4.1",
     "cosmiconfig": "^5.0.0",
-    "create-jest-runner": "^0.5.3",
-    "eslint": "^5.6.0"
```
kentcdodds: Having eslint listed in the dependencies resulted in potentially downloading multiple copies of eslint. Let's rely on the peerDep people should be installing themselves.

ljharb:
It should; it's totally fine (and often preferred) to have a peerDep listed also as a dep - regardless, peer deps should *always* be listed as either peer + dep, or peer + dev dep - *never* just as a peer dep.

package.json
ljharb: deduping handles this - it's supposed to be both.

kentcdodds: No, deduping will not handle this. It will see that this module requires no greater than eslint 5 and my project requires eslint 6 so it will install both.

ljharb: oh sure, that's the issue with what's on master - that's why the RHS of "eslint" must match the peer dep :-)

kentcdodds: Sorry, I'm not sure I understand. So are your saying the PR is good?

ljharb: No - I'm saying that since the peerDep string is `"^4.0.0 || ^5.0.0 || ^6.0.0"`, `eslint` needs the same exact value in either deps or devDeps.

Putting it in deps means that people don't have to necessarily have eslint installed to invoke it via this runner, which is the goal here. If there's an incompatible eslint installed up the tree, the peer + dep combo will make `npm ls` fail (which everyone should be always running in CI anyways, since if it fails, your dep graph is invalid and you can't rely on anything working)

kentcdodds: I see what you mean. It's wired because if someone doesn't have eslint installed they'll get a warning in the terminal, but it's your project, so sure.

ljharb: With peer + dep, they won't in fact get a warning - eslint will just be automatically installed and everything will Just Work.

ljharb:
Looks like tests are failing; more test changes may be required here.

kentcdodds:
I don't have a lot of bandwidth here 😬, but maybe changing the `parserOptions` from `null` to `{}` will fix it.

SimenB:
`eslint@6` has dropped support for node 6, so that will keep failing until we drop it as well.

---

I wonder if we should run CI against all versions of eslint we claim to support. Right now we only run against whatever is the latest of them.

ljharb:
@SimenB we can just exclude the eslint 6 / node 6 combo. We definitely should be testing every eslint and node version we support together.

SimenB:
Yeah, that makes sense to me. Last time I had a testing matrix like that I gave up on travis and moved to circle ci, but I bet we can make it work now that travis has stages and other more advanced options

ljharb:
@SimenB it's pretty easy with travis, see the [airbnb config monorepo](https://github.com/airbnb/javascript/blob/master/.travis.yml)

PR #76, opened 2019-09-21
fix: parserOptions, rules default values for ESLint 6

wheeler:
# Problem
When trying to use `jest-runner-eslint` on a project being upgraded to ESLint 6 there are the following errors:
```
ESLint configuration in CLIOptions is invalid:
    - Property "parserOptions" is the wrong type (expected object but got `null`).
    - Property "rules" is the wrong type (expected object but got `null`).
```

### Setup
I "use ESLint 6" by forcing the version using the following in my `package.json`
```json
  "resolutions": { "eslint": "6.4.0" },
```

# Cause
ESLint 6 introduces type checks on the config.
See ~ **PR:** https://github.com/eslint/eslint/pull/11546 || **Commit:** https://github.com/eslint/eslint/commit/6ae21a4bfe5a1566f787fbad798182a524b96d28

Two default values set in `normalizeConfig.js` are not compatible.

- `config.parserOptions` is typed as an `object` so `null` is not valid.
https://github.com/eslint/eslint/blob/2e202ca2228846e6226aa8dd99c614d572fb86a8/lib/shared/types.js#L22

- `config.rules` is typed as a `Record` so `null` is not valid.
https://github.com/eslint/eslint/blob/2e202ca2228846e6226aa8dd99c614d572fb86a8/lib/shared/types.js#L41

# Fix

Change the default values for these two configurations from `null` to `{}` in `normalizeConfig.js`.

Using this change `jest-runner-eslint` + ESlint 6 runs normally for my project. I also tested with this change but back on ESLint 5 and that continues to work.

This is not intended to fully support ESLint 6 but step in the right direction.

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,5 +220,5 @@ jest-runner-eslint maps a lot of ESLint CLI arguments to config options. For exa
 | plugin                        | `[]`           | `"plugin": "prettier"` or `"plugin": ["pettier", "other"]`                                    |
 | quiet                         | `true`         | `"quiet": false`                                                                              |
 | reportUnusedDisableDirectives | `false`        | `"reportUnusedDisableDirectives": true`                                                       |
-| rules                         | `null`         | `"rules": {"quotes": [2, "double"]}` or `"rules": {"quotes": [2, "double"], "no-console": 2}` |
+| rules                         | `{}`           | `"rules": {"quotes": [2, "double"]}` or `"rules": {"quotes": [2, "double"], "no-console": 2}` |
 | rulesdir                      | `[]`           | `"rulesdir": "/path/to/rules/dir"` or `"env": ["/path/to/rules/dir", "/path/to/other"]`       |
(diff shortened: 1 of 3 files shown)


ljharb:
This seems like an overlap of #70.

wheeler:
@ljharb I agree there is overlap. #70 fixes one of these fields under the umbrella of allowing ESLint. This fixes two. This PR does not contain the assertion that ESLint 6 is fully supported - but it does not have any impact on ESLint 5. If this were merged and #70 were rebased it would then contain only the `package.json` changes allowing ESLint 6 (and possibly be closer to passing).

ljharb:
There's no value imo in adding anything less than full support for ES6 (which is sorely needed)

wheeler:
OK, closing.

wheeler:
@ljharb more complete version here -> #77

Issue #75, opened 2019-09-16
Support ESLint 6


tunnckoCore:
Seems like it fails when we use `overrides`

```
  ● Test suite failed to run

    ESLint configuration in /home/charlike/github/tunnckoCore/configs/.eslintrc.js is invalid:
        - Unexpected top-level property "overrides[0].extends".

      at validateConfigSchema (node_modules/jest-runner-eslint/node_modules/eslint/lib/config/conf
ig-validator.js:235:15)
```

And that's part of my eslint config

```js
  overrides: [
    {
      files: [
        '**/__tests__/**',
        '**/tests/**',
        '**/test/**',
        '**/*.{test,spec}.{js,jsx,ts,tsx,mdx}',
      ],
      extends: ['plugin:jest/recommended'],
    },
    {
      files: ['**/*.ts', '**/*.tsx'],
      parser: '@typescript-eslint/parser',
      parserOptions: {
        sourceType: 'module',
        jsx: true,
        project: 'tsconfig.json',
      },
      extends: [
        'plugin:@typescript-eslint/eslint-recommended',
        'plugin:@typescript-eslint/recommended',
        'plugin:@typescript-eslint/recommended-requiring-type-checking',

        // TODO @tunnckoCore [2200-12-12]: Consider xo-typescript
        // 'xo-typescript',

        // TODO @tunnckoCore [2200-12-12]: Do we need Prettier here too? Check how overrides works!
        'plugin:prettier/recommended',
        'prettier/@typescript-eslint',
      ],
      rules: {
        // Enforce both my and Airbnb's preference used for years in gazillion ton of code
        '@typescript-eslint/no-unused-vars': 'error',
        // Inferring is a good thing to some extent.
        '@typescript-eslint/explicit-function-return-type': 'off',
        'no-use-before-define': 'off',
        '@typescript-eslint/no-use-before-define': [
          'error',
          {
            functions: false,
            classes: true,
            variables: true,
            typedefs: true,
          },
        ],
      },
    },
    {
      // TODO @tunnckoCore [2200-12-12]: eslint errors with "uknown jsx node"
      files: ['**/*.mdx'],
      parser: 'eslint-mdx',
      plugins: ['mdx'],
      rules: {
        'mdx/no-jsx-html-comments': 'error',
        'mdx/no-unescaped-entities': 'warn',
        'mdx/no-unused-expressions': 'error',
        'mdx/remark': 'warn',
        'no-unused-expressions': 'off',
[...]

PR #77, opened 2019-09-21
Allow ESLint6

wheeler:
This PR combines #70 and #76, then continues by resolving additional test failures and configuring Travis testing matrix across ESLint versions.

#### Changes:
- [x] (from #70) expand ESLint deps `^5.6.0` ▶️ `^4.0.0 || ^5.0.0 || ^6.0.0`
- [x] (from #70 & #76) update defaults for parserOptions & rules (to satisfy type definitions described in #76)
- [x] (from #76) update README.md
- [x] `integrationTests/custom-parser.test.js > Doesn't override parser when not set` receives different text for the same result from ESLint6. Change the test to expect the result according to the version being used†.
- [x] config travis to run a `node` * `eslint` version matrix, skipping the combination `node@6 + eslint@6`
- [x] update the jest logo url in README.md
- [x] [**CNR**] try to reproduce the problem reported in issue #75

† This was done with two special techniques: 1) checking `eslint.CLIEngine.version`, 2) adding a snapshot for both versions and "skipping" the version not being tested. Both techniques have multiple alternate solutions. ex. 1) could be made to import the `package.json` file and check version there. 2) could be converted to inline snapshots. Open to alternate solutions if the current ones aren't palatable.

Closes #70

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,14 @@ node_js:
   - '10'
   - '8'
   - '6'
+install:
+  - 'if [ -n "${ESLINT}" ]; then yarn add "eslint@${ESLINT}"; fi'
+  - 'yarn'
+env:
+  - 'ESLINT=6'
+  - 'ESLINT=5'
+  - 'ESLINT=4'
+matrix:
+  exclude:
+    - node_js: '6'
+      env: 'ESLINT=6'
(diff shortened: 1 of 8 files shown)


wheeler:
poke @ljharb - any thoughts on this more complete ESLint6 PR?

SimenB:
https://github.com/jest-community/jest-runner-eslint/releases/tag/v0.7.5

PR #79, opened 2019-10-25
chore: docs - correct default value for ‘quiet’

wheeler:
Fixing this default value in the docs which doesn't match what it is in normalizeConfig.js


wheeler:
@ljharb note: I can't merge this.

PR #83, opened 2020-01-22
chore: upgrade jest to v25

grxy:


grxy:
@ljharb Thx for the review. Looks like we'll need to remove node 6 support and update or remove node 8 support as well. :/

grxy:
@ljharb That sounds good. I'll model it after how it's done for `eslint` versions.

grxy:
@ljharb After adding a full eslint x jest matrix, it turns out there are some dependency issues with older versions of jest (See https://travis-ci.org/jest-community/jest-runner-eslint/builds/640649231). Perhaps it would be more useful to the community to get a version out there that supports jest 25 (and dropping node 6/8 CI coverage temporarily), then following up with a deeper dive into a full jest x eslint CI matrix. What do you think? Or do you have a better approach to solving the matrix issue?

.travis.yml
```diff
@@ -5,13 +5,49 @@ node_js:
   - '8'
   - '6'
 install:
+  - 'if [ -n "${JEST}" ]; then yarn add --dev "jest@${JEST}"; fi'
```
SimenB:
```suggestion
  - 'if [ -n "${JEST}" ]; then yarn add --dev "jest@${JEST}" "babel-jest@${JEST}"; fi'
```

Not sure how that cooperates with babel 6 vs 7, though... I'm unsure if it's feasible to test with more than 24 and 25, as those are the only versions that support babel 7 out of the box. With 22 and 23 we'll also need `babel-core@bridge` (not a biggie, so I guess you can try adding it as well). Jest 21 doesn't support babel 7 at all, so that's a no-go

.travis.yml
```diff
@@ -5,13 +5,49 @@ node_js:
...
+    - node_js: '6'
+      env: 'ESLINT=6 JEST=25'
+    # exclude jest 25 on node 6
+    - node_js: '6'
+      env: 'ESLINT=4 JEST=25'
+    - node_js: '6'
+      env: 'ESLINT=5 JEST=25'
+    - node_js: '6'
+      env: 'ESLINT=6 JEST=25'
+    # exclude jest 25 on node 8
```
SimenB: why? Jest 25 still works with node 8

package.json
```diff
@@ -40,15 +40,15 @@
     "eslint-plugin-jest": "^22.6.4",
     "eslint-plugin-prettier": "^3.1.0",
     "execa": "^1.0.0",
-    "jest": "^21.0.0 || ^22.0.0 || ^23.0.0 || ^24.0.0",
+    "jest": "^21.0.0 || ^22.0.0 || ^23.0.0 || ^24.0.0 || ^25.0.0",
```
SimenB:
```suggestion
    "jest": "^21.0.0 || ^22.0.0 || ^23.0.0 || ^24.0.0 || ^25.1.0",
```

25.0.0 is a beta version released back in the summer, unfortunately as version `25.0.0` (never tagged as `latest` though). 25.1.0 is the first stable release of 25

package.json
```diff
@@ -40,15 +40,15 @@
...
     "jest-watch-select-projects": "^0.1.1",
     "jest-watch-typeahead": "^0.3.1",
     "prettier": "^1.14.3",
     "rimraf": "^2.6.2"
   },
   "peerDependencies": {
     "eslint": "^4.0.0 || ^5.0.0 || ^6.0.0",
-    "jest": "^21.0.0 || ^22.0.0 || ^23.0.0 || ^24.0.0"
+    "jest": "^21.0.0 || ^22.0.0 || ^23.0.0 || ^24.0.0 || ^25.0.0"
```
SimenB:
```suggestion
    "jest": "^21.0.0 || ^22.0.0 || ^23.0.0 || ^24.0.0 || ^25.1.0"
```

package.json
grxy: Good to know! Will update.

.travis.yml
grxy: This is because jest-snapshot v25 requires a minimum of semver 7.1.1, which requires node 10:

image

ljharb: Then that’s probably a bug in jest-snapshot, which should still support node 8 whole jest does?

grxy: Yeah I would think so, but not sure if that should block getting a version of this landed that supports jest 25. What do you think? Happy to file an issue.

ljharb: Agree it shouldn’t block for now; filing a jest issue would be great too.

grxy: Issue filed here: https://github.com/facebook/jest/issues/9450

.travis.yml
grxy: I've updated the build matrix to run jest 24-25 x eslint 4-6 x node 6-12 for now. PR should be in a good place unless you have other suggestions.

.travis.yml
```diff
@@ -5,8 +5,7 @@ node_js:
   - '8'
   - '6'
 install:
-  - 'if [ -n "${JEST}" ]; then yarn add --dev "jest@${JEST}" "babel-jest@${JEST}"; fi'
-  - 'if [ -n "${ESLINT}" ]; then yarn add "eslint@${ESLINT}"; fi'
+  - 'yarn add "eslint@${ESLINT}" "jest@${JEST}" "babel-jest@${JEST}"'
```
ljharb: if this isn’t done conditionally, won’t it trigger on a future, say, linting build or similar? I suppose we could add the condition then too tho.

ljharb: Most projects I've seen have an explicit `include` that runs eslint, since you don't need to run that more than once. My projects also have one for coverage, and one for `npx aud` for an audit check. This project doesn't seem to have that yet, however.

.travis.yml
SimenB: merged it, but I'm not sure when we'll get a patch release out. Omitting it for now in the testing matrix here makes sense

grxy:
@SimenB @ljharb Do either of you have any other changes you'd like me to make, or are we able to land this and get a new release out there?

SimenB:
I just left my apartment, but I'll make a release when I get back home in a couple of hours. I'll also add @ljharb to the npm package, seems we've forgotten to do that 😬

grxy:
Thanks, y'all! 🙏

SimenB:
`0.7.6` released, and @ljharb added to the package.

Thanks @grxy!

PR #85, opened 2020-01-27
chore: refresh lockfiles to get rid of security alert

SimenB:
It just complains about `set-value` which we get via `chokidar`, `sane` and `micromatch` - all development dependencies.

I just deleted the lockfiles and ran `yarn` again. Will merge if green


ljharb:
seems simpler to get rid of the lockfiles for the examples.

SimenB:
Yeah, let's

SimenB:
`@babel/preset-env` depends on `levenary`, which is node 8+...

SimenB:
I can force a lower version of `preset-env` to unblock. It's just a dev tool, so doesn't affect consumers

ljharb:
That seems like a breaking change for preset-env; could you file an issue?

SimenB:
Yeah, I added a comment to the PR which added it to babel

SimenB:
Upstream fix: https://github.com/tanhauhau/levenary/pull/3

If merged and released we can upgrade again.

PR #88, opened 2020-04-10
fix: reuse CLIEngine instances

### LINKED ISSUE #87: Reuse CLIEngine instances
Right now the runner will instantiate a new CLIEngine instance for every file it runs on. 

https://github.com/jest-community/jest-runner-eslint/blob/a239b614bce32f14d4dac88209fe0d9eacfca6dc/src/runner/runESLint.js#L20-L27

I think we can make this more efficient by reusing the same instance every time. To try this out locally, I just hacked this up in my node_modules:

```diff
30a31,46
> let cachedCli;
> function getCli(config, extraOptions) {
>   if (!cachedCli) {
>     const _getESLintOptions = getESLintOptions(config),
>           baseCliOptions = _getESLintOptions.cliOptions;
> 
>     const cliOptions = _objectSpread({}, baseCliOptions, {
>       fix: getComputedFixValue(baseCliOptions)
>     }, extraOptions);
> 
>     cachedCli = { cliOptions, cli: new CLIEngine(cliOptions) };
>   }
> 
>   return cachedCli;
> }
> 
43,50c59
<   const _getESLintOptions = getESLintOptions(config),
<         baseCliOptions = _getESLintOptions.cliOptions;
< 
<   const cliOptions = _objectSpread({}, baseCliOptions, {
<     fix: getComputedFixValue(baseCliOptions)
<   }, extraOptions);
< 
<   const cli = new CLIEngine(cliOptions);
---
>   const { cli, cliOptions } = getCli(config, extraOptions);
```

My very non-scientific n=1 run on 761 files before this change:

574.30s user 86.33s system 1029% cpu 1:04.16 total

And after:

534.90s user 81.40s system 969% cpu 1:03.55 total

I'm not sure if the config or extraOptions can change between executions of this function, but it seems like they should be stable. If not, we'd want to do some smarter memoization here.

Thoughts?

lencioni:
Right now the runner will instantiate a new CLIEngine instance for every
file it runs on. We can make this more efficient by reusing the same
instance every time. This will make it the CLIEngine is instantiated
once per worker instead of once per file.

My very non-scientific n=1 run on 761 files before this change:

574.30s user 86.33s system 1029% cpu 1:04.16 total

And after:

534.90s user 81.40s system 969% cpu 1:03.55 total

I verified that CLIEngine can be reused on files in different
directories while respecting nested eslintrc files.

Here's my test script:

```js
const { CLIEngine } = require('eslint');

const cli = new CLIEngine({});

const paths = ['./toplevel.js', './nested/nested.js'];
paths.forEach(path => {
  const report = cli.executeOnFiles(require.resolve(path));
  console.log(report.results);
});
```

In the two files to be linted, I put this:

```js
foo();
```

And in the eslintrc files I set this at the top level:

```json
{ "rules": { "no-undef": "error" } }
```

And this in the nested directory:

```json
{ "rules": { "no-undef": "warn" } }
```

Here's what was logged:

```js
[ { filePath: '/Users/joe_lencioni/cliengine-test/toplevel.js',
    messages: [ [Object] ],
    errorCount: 1,
    warningCount: 0,
    fixableErrorCount: 0,
    fixableWarningCount: 0,
    source: 'foo();\n' } ]
[ { filePath: '/Users/joe_lencioni/cliengine-test/nested/nested.js',
    messages: [ [Object] ],
    errorCount: 0,
    warningCount: 1,
    fixableErrorCount: 0,
    fixableWarningCount: 0,
    source: 'foo();\n' } ]
```

Fixes #87

### DIFF
--- a/src/runner/runESLint.js
+++ b/src/runner/runESLint.js
@@ -9,6 +9,24 @@ const getComputedFixValue = ({ fix, quiet, fixDryRun }) => {
   return undefined;
 };
 
+let cachedValues;
+const getCachedValues = (config, extraOptions) => {
+  if (!cachedValues) {
+    const { cliOptions: baseCliOptions } = getESLintOptions(config);
+    const cliOptions = {
+      ...baseCliOptions,
+      fix: getComputedFixValue(baseCliOptions),
+      ...extraOptions,
+    };
+    const cli = new CLIEngine(cliOptions);
+    const formatter = cli.getFormatter(cliOptions.format);
+
+    cachedValues = { cli, formatter, cliOptions };
+  }
+
+  return cachedValues;
+};
+
 const runESLint = ({ testPath, config, extraOptions }) => {
   const start = Date.now();
 
@@ -17,14 +35,7 @@ const runESLint = ({ testPath, config, extraOptions }) => {
     require(config.setupTestFrameworkScriptFile);
   }
 
-  const { cliOptions: baseCliOptions } = getESLintOptions(config);
-  const cliOptions = {
-    ...baseCliOptions,
-    fix: getComputedFixValue(baseCliOptions),
-    ...extraOptions,
-  };
-
-  const cli = new CLIEngine(cliOptions);
+  const { cli, formatter, cliOptions } = getCachedValues(config, extraOptions);
 
   if (cli.isPathIgnored(testPath)) {
     const end = Date.now();
@@ -39,7 +50,7 @@ const runESLint = ({ testPath, config, extraOptions }) => {
 
   const end = Date.now();
 
-  const message = cli.getFormatter(cliOptions.format)(
+  const message = formatter(
     cliOptions.quiet
       ? CLIEngine.getErrorResults(report.results)
       : report.results,


src/runner/runESLint.js
```diff
@@ -17,14 +35,7 @@ const runESLint = ({ testPath, config, extraOptions }) => {
...
 
-  const { cliOptions: baseCliOptions } = getESLintOptions(config);
-  const cliOptions = {
-    ...baseCliOptions,
-    fix: getComputedFixValue(baseCliOptions),
-    ...extraOptions,
-  };
-
-  const cli = new CLIEngine(cliOptions);
+  const { cli, formatter, cliOptions } = getCachedValues(config, extraOptions);
```
ljharb, line 38: what happens if runESLint is called with a different config/extraOptions when cached values exist? Is that worth checking for?

lencioni: Possibly, but I'm not sure realistically if this would ever happen, since I don't think you can modify the config while it is running. Maybe watch mode? Even then, I'm not sure it would be worth the overhead.

lencioni:
@ljharb Do you want me to make any changes to this PR before merging?

lencioni:
@SimenB When you have a moment, could you please review this change? 😃  Thank you! 🙏

lencioni:
@SimenB @rogeliog Any chance you could take a look at this PR? Thank you!

lencioni:
Thanks @SimenB! I've verified locally by adding some logging to my jest runner that the workers are not reused across watch mode runs, so that caching gets wiped out anyway, so I think we are good to go here.

SimenB:
CI is green, not sure why it's still "pending" here...
image

SimenB:
minor or path?

ljharb:
Seems like a patch, since it shouldn't likely be observable.

SimenB:
Sounds good, 0.7.7 released 👍

PR #90, opened 2020-05-05
feat: support jest v26

Moeriki:


package.json
```diff
@@ -40,15 +40,15 @@
     "eslint-plugin-jest": "^22.6.4",
     "eslint-plugin-prettier": "^3.1.0",
     "execa": "^1.0.0",
-    "jest": "^21.0.0 || ^22.0.0 || ^23.0.0 || ^24.0.0 || ^25.1.0",
+    "jest": "^26.0.0",
```
ljharb, line 43: This line needs to exactly match the peerdep.

Moeriki: Ah sorry thought it was an oversight. Doesn't npm just install the latest?

Moeriki:
It seems the snapshot `normalize` isn't effective anymore. I get snapshot failures because of test duration. I'm still checking why.

package.json
ljharb: If none is otherwise installed, yes; but in tests, we'll want to install each supported version, and `npm ls` should pass there.

Moeriki:
Had to update

```diff
- /\(?\d*\.?\d+m?s\)?/g
+ /\(?\d*\.?\d+ m?s\b\)?/g
```

Space because jest updated its duration formating, and `\b` because otherwise it would match with strings like `1 skipped`.

integrationTests/__snapshots__/custom-parser.test.js.snap
```diff
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Doesn't override parser when not set 1`] = `
+exports[`Doesn't override parser when not set [ESLint<6] 1`] = `
```
SimenB, line 3: seems like this is failing on CI?

.travis.yml
```diff
@@ -35,3 +38,10 @@ matrix:
       env: 'ESLINT=5 JEST=25'
     - node_js: '8'
```
SimenB, line 39: we can run jest 25 on node 8, the bug was fixed. not really for this PR though 🙂

.travis.yml
```diff
@@ -35,3 +38,10 @@ matrix:
...
       env: 'ESLINT=6 JEST=25'
+    # exclude jest 26 on node 8
```
SimenB, line 41: need to skip it on node 6 as well

integrationTests/__snapshots__/custom-parser.test.js.snap
Moeriki: The snapshot was locally updated by jest 26. It seems they strip it for a reason. I'm checking the release notes.

.travis.yml
Moeriki: Ah I see. I was going to ask :)

.travis.yml
Moeriki: Fixed.

Moeriki:
I think I misunderstood something in the testing setup.

Had some failures and updated some snapshots, while I shouldn't have.

All seems good now.

integrationTests/__snapshots__/custom-parser.test.js.snap
Moeriki: Nevermind I did something wrong. They all pass!

SimenB:
`v0.8.0` published, thanks @Moeriki!

PR #78, opened 2019-10-10
chore: bump eslint-config-airbnb

SimenB:
Gets rid of peer dependency warning when running `yarn`. Also bumps `set-value` and `handlebars` for vulnerability fixes

PR #92, opened 2020-05-17
feat: Support ESLint 7.x

### LINKED ISSUE #91: Support ESLint 7.x
ESLint v7.0.0 is [released](https://eslint.org/blog/2020/05/eslint-v7.0.0-released) 🎉

It would be awesome to have official ESLint 7 support. 👊
I'm happy to help where I can of course 🙂

MichaelDeBoey:
ESLint v7.0.0 is [released](https://eslint.org/blog/2020/05/eslint-v7.0.0-released) 🎉

(dev)Dependencies should be compatible with ESLint 7 too before we can merge this one:

- [x] [`eslint-config-airbnb-base`](https://github.com/airbnb/javascript/issues/2223)
  - [x]  [PR](https://github.com/airbnb/javascript/pull/2240)
  - [x]  [Release](https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/CHANGELOG.md#1420--2020-06-10)
- [x] [`eslint-plugin-import`](https://github.com/benmosher/eslint-plugin-import/issues/1772)
  - [x]  [PR](https://github.com/benmosher/eslint-plugin-import/pull/1715) + [commit](https://github.com/benmosher/eslint-plugin-import/commit/6a110dd16a7fd775f08601054bf14ffd503eea7b)
  - [x]  [Release](https://github.com/benmosher/eslint-plugin-import/blob/master/CHANGELOG.md#2211---2020-06-07)
- [x] [`eslint-plugin-jest`](https://github.com/jest-community/eslint-plugin-jest/issues/568)
  - [x]  [PR](https://github.com/jest-community/eslint-plugin-jest/pull/572)
  - [x]  Released
- [x] [`eslint-plugin-prettier`](https://github.com/prettier/eslint-plugin-prettier/issues/292)
  - [x]  [PR](https://github.com/prettier/eslint-plugin-prettier/issues/291)
  - [x]  Released

---

Closes #91


.travis.yml
```diff
@@ -1,54 +1,72 @@
...
 node_js:
-  - '12'
-  - '10'
-  - '8'
-  - '6'
+  - 6
+  - 8
+  - 10
+  - 12
+  - 14
```
ljharb, line 7:
```suggestion
  - 14
  - 12
  - 10
  - 8
  - 6
```
nit: descending

ljharb:
@MichaelDeBoey please rebase and mark as "ready for review" once tests are passing :-)

MichaelDeBoey:
@ljharb This one's waiting for `eslint-plugin-import` to be merged

ljharb:
aha, thanks, that’s on me. Hopefully this week.

SimenB:
CI seems unhappy for the same reason as in #78. We probably want to leave the eslint config alone

package.json
```diff
@@ -27,18 +27,18 @@
...
     "@babel/cli": "^7.1.0",
     "@babel/core": "^7.1.0",
     "@babel/preset-env": "~7.7.0",
     "babel-jest": "^25.1.0",
-    "eslint-config-airbnb-base": "^13.1.0",
-    "eslint-config-prettier": "^4.3.0",
-    "eslint-plugin-import": "^2.14.0",
-    "eslint-plugin-jest": "^22.6.4",
-    "eslint-plugin-prettier": "^3.1.0",
+    "eslint-config-airbnb-base": "^14.1.0",
```
SimenB, line 37: I _think_ this is why CI fails - keep it at 13?

ljharb: No; the airbnb config does not yet support eslint 7. Thus, this PR is blocked until https://github.com/airbnb/javascript/issues/2223 is resolved.

MichaelDeBoey:
@ljharb Any idea how to resolve ESLint errors?

ljharb:
I see a few errors; one of them is because yarn incorrectly fails installs when engines mismatch (they’re advisory only); if there’s a way to yarn install without that check, that’s the fix there (or switching to the standard package manager, of course)

The linting error suggests that perhaps eslint isn’t properly updated in the lockfile? I don’t use lockfiles on packages so I’m not sure, but my guess is if you deleted and recreated the lockfile it’d pass.

ljharb:
Seems like the eslint 4 tests won't work with the camelcase rule; i wonder if we could override those, with something like `--rule=camelcase:0` or something?

SimenB:
`yarn --ignore-engines`

ljharb:
Looks like the `yarn add` commands also need `--ignore-engines`.

ljharb:
k i think the last issue is that for eslint 4, we have to have a way to override the config for the `camelcase` rule, since it has options that only exist in eslint 5+.

MichaelDeBoey:
@ljharb Any idea how I could do this override?

ljharb:
You could pass `--rule="camelcase:0"` in travis.yml to disable it, for example?

MichaelDeBoey:
@ljharb I tried just adding it, but that didn't work.
I also have no clue how I should do it otherwise, so I would be happy if you could help me out here.

MichaelDeBoey:
@ljharb Any idea how I could resolve the failures exactly?
Would love to have this one (and all other ESLint PRs) merged asap 🙂

lencioni:
I think we should just drop support for eslint 4 here.

MichaelDeBoey:
@lencioni I'm happy with that too.

@ljharb @SimenB Opinion on dropping Node 4 support?

SimenB:
I'm down with dropping both eslint 4 and node 4. Up to @ljharb

MichaelDeBoey:
I've made the changes locally, so can commit whenever you're down too @ljharb 🙂

ljharb:
I'd prefer not to drop them if possible :-)

I'll take a crack at getting the tests to pass later tonight; if I can't make it work, then we can update this PR to be semver-major tomorrow.

lencioni:
This package is still pre-1.0 so that change could be either 0.10.0 or 1.0.0.

ljharb:
First, I updated the test matrix; the peer deps support jest 21-26 and eslint 4-7, and the implied node version somewhere between 4-14, depending on which combo you're using.

As is, tests are failing on jest 21, 22, and 23 on master, as well as on eslint 4 with all versions of jest: https://travis-ci.org/github/jest-community/jest-runner-eslint/builds/704554241

---
Next, I tried to fix the eslint 4 failures, which were just about the camelcase rule's extra option. I did that by, in travis, when eslint was v4, installing eslint-config-airbnb-base@13. That left these failures: https://travis-ci.org/github/jest-community/jest-runner-eslint/builds/704558140 which are, "still failing on jest < 24" but passing on eslint 4.

---
Failures are: on jest 21, "TypeError: glob pattern string required"; on jest 22, "SecurityError: localStorage is not available for opaque origins" (fixed with advice from https://github.com/jsdom/jsdom/issues/2304); and on jest 22 and 23, related to babel 6 vs 7 : https://travis-ci.org/github/jest-community/jest-runner-eslint/builds/704562583

---
The node 4 failure is because travis' default yarn version doesn't work on node 4, yay. 


In other words, it seems like unless we want to a) ditch yarn (which i strongly recommend anyways, for a variety of reasons), b) do some babel 6/7 hijinx in travis.yml, c) fix whatever *actual* test failures pop up, if any (i actually suspect everything works fine! it's just that it hasn't been properly tested for a long time, so who knows) - then we'd want to just drop whatever's failing.

It also looks like "projects" isn't supported in jest 21 at all inside a config file.

So, I added a commit that drops support for eslint < 6 and jest < 25 (ie, so we keep supporting the last 2 versions of both eslint and jest). Moving forward, we'll be able to ensure we test every node and eslint and jest version combo that's supported, and prevent this kind of situation from happening again.

SimenB:
Supporting last 2 majors seems reasonable to me! 👍

.travis.yml
```diff
@@ -1,54 +1,21 @@
...
-  - '6'
+  # eslint 7: node 10+
+  # jest 26: node 10+
+  - 14
+  - 12
+  - 10
 install:
-  - 'yarn add "eslint@${ESLINT}" "jest@${JEST}" "babel-jest@${JEST}"'
-  - 'yarn'
+  - 'yarn add "eslint@${ESLINT}" "jest@${JEST}" "babel-jest@${JEST}" --ignore-engines'
```
SimenB, line 9: we shouldn't need `ignore-engines` anymore, but I guess it doesn't hurt either

MichaelDeBoey: Removed 🙂

MichaelDeBoey:
CI failed because of removing `--ignore-engines`, so I added it back:
https://travis-ci.org/github/jest-community/jest-runner-eslint/jobs/704613630

SimenB:
Huh - what's installing semver@7 when we target node 8?

MichaelDeBoey:
Investigating as we speak 🙂

MichaelDeBoey:
@SimenB It seems to be triggered whenever `eslint@6` is installed.
However when I create a clean project and I install `eslint@6`, it will install `semver@6.3.0`
Schermafbeelding 2020-07-03 om 12 37 51

package.json
```diff
@@ -54,5 +54,8 @@
     "proseWrap": "never",
     "singleQuote": true,
     "trailingComma": "all"
+  },
+  "engines": {
+    "node": ">=8.10.0"
   }
```
ljharb, line 60: good catch

ljharb:
(note that `--ignore-engines` should be the default; the "engines" field is and always has been purely advisory, so i think it's fine to not resolve this yarn error)

MichaelDeBoey:
@ljharb `--ignore-engines` is added again, so this is good to be merged 🙂

MichaelDeBoey:
Thanks for merging @ljharb! 👊
Any idea when this will be released? 🤔

.travis.yml
```diff
@@ -1,54 +1,21 @@
 language: node_js
 node_js:
-  - '12'
-  - '10'
-  - '8'
```
SimenB, line 5: why aren't we testing on 8 if 8 is in our `engines`?

MichaelDeBoey: We do.

But instead of ignoring, we only include the only test case for Node 8

SimenB: aha! It's been so long since I used travis I've forgotten how the config works 😅

MichaelDeBoey:
@SimenB Any idea when this will be released? 🤔

SimenB:
I'd like to bump all other dependencies as well first (at least the ones that still support node 8)

MichaelDeBoey:
@SimenB On it boss! 😉

SimenB:
@MichaelDeBoey I've opened up https://github.com/jest-community/jest-runner-eslint/pull/98

SimenB:
@MichaelDeBoey available in https://github.com/jest-community/jest-runner-eslint/releases/tag/v0.10.0

PR #95, opened 2020-05-26
feat(cli-options): support resolvePluginsRelativeTo

### LINKED ISSUE #94: CLI options: support --resolve-plugins-relative-to
See [ESLint docs](https://eslint.org/docs/user-guide/command-line-interface#resolve-plugins-relative-to) for the option.
This should be a rather simple change [here](https://github.com/jest-community/jest-runner-eslint/blob/4b4d99cc12cf2d870ff45d6eb4d56920d6d9359d/src/utils/normalizeConfig.js).

jeysal:
Closes #94.

I did not add a test case to `normalizeConfig.test.js` as it already doesn't have test cases for every option, only for every option type to make sure the type conversions work. This is pretty declarative anyway so declaring the new option again in the test would just be repitition.
Manually tested by linking into a project that does not work without setting the option - it works now.

Could consider passing the cwd to ESLint ourselves as a default instead of relying on the ESLint default by passing `undefined` as a default. This is the first option to pass `undefined` by default. But I don't see a reason why `jest-runner-eslint` would need to take control of the default option here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,7 +218,8 @@ jest-runner-eslint maps a lot of ESLint CLI arguments to config options. For exa
 | parser                        | `espree`       | `"parser": "flow"`                                                                            |
 | parserOptions                 | `{}`           | `"parserOptions": { "myOption": true }`                                                       |
 | plugin                        | `[]`           | `"plugin": "prettier"` or `"plugin": ["pettier", "other"]`                                    |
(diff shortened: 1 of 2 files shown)


SimenB:
Passing `rootDir` as `cwd` might make sense, dunno

SimenB:
Published as 0.9.0

PR #97, opened 2020-07-04
fix: Fix dependencies

MichaelDeBoey:
- ESLint can't be a dependency + peerDependency
- `babel-jest` should have the same restriction like `jest` (like we do in our tests)

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,14 +26,14 @@
   "dependencies": {
     "chalk": "^2.4.1",
     "cosmiconfig": "^5.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -26,14 +26,14 @@
   "dependencies": {
     "chalk": "^2.4.1",
     "cosmiconfig": "^5.0.0",
-    "create-jest-runner": "^0.5.3",
-    "eslint": "^6 || ^7"
```
SimenB, line 30: good catch!

PR #99, opened 2020-07-04
chore: Update dependencies

MichaelDeBoey:


package.json
```diff
@@ -29,21 +29,21 @@
...
+    "@babel/preset-env": "^7.10.4",
     "babel-jest": "^25.1 || ^26",
     "eslint": "^6 || ^7",
     "eslint-config-airbnb-base": "^14.2.0",
     "eslint-config-prettier": "^6.11.0",
     "eslint-plugin-import": "^2.22.0",
     "eslint-plugin-jest": "^23.17.1",
     "eslint-plugin-prettier": "^3.1.4",
-    "execa": "^2.1.0",
+    "execa": "^3.4.0",
```
SimenB, line 42: for some reason I though v3 dropped node 8. seems not!

package.json
```diff
@@ -29,21 +29,21 @@
...
     "jest": "^25.1 || ^26",
     "jest-watch-select-projects": "^2.0.0",
     "jest-watch-typeahead": "^0.5.0",
-    "prettier": "^1.14.3",
+    "prettier": "1.19.1",
```
SimenB, line 46: we don't need to do this since we have a lockfile. doesn't hurt either, tho

package.json
```diff
@@ -29,21 +29,21 @@
     "create-jest-runner": "^0.6.0"
   },
   "devDependencies": {
-    "@babel/cli": "^7.1.0",
-    "@babel/core": "^7.1.0",
-    "@babel/preset-env": "^7.1.0",
+    "@babel/cli": "^7.10.4",
```
SimenB, line 32: ?

MichaelDeBoey: Just to make it the highest possible, that's all

package.json
MichaelDeBoey: Nope v4 drops Node 8

SimenB:
Thanks!

PR #106, opened 2021-04-06
Use latest `create-jest-runner` dependency

mrtnzlml:


ljharb (approved):

CONTRIBUTING.md
```diff
@@ -6,7 +6,7 @@
 
    ```sh
    git clone https://github.com/<your_username>/jest-runner-eslint
-   cd jest
+   cd jest-runner-eslint
```
ljharb, line 9: this is a good change regardless

ljharb (changes requested):
Actually, it looks like tests are failing in node 8, because of syntax node 8 doesn't support, so it doesn't seem like we can update to this version without a breaking change.

mrtnzlml:
> Actually, it looks like tests are failing in node 8, because of syntax node 8 doesn't support, so it doesn't seem like we can update to this version without a breaking change.

@ljharb Is there anything preventing us to do so? I see that this project didn't reach a stable version yet (it's 0.10.0) so I could just send another PR. But not sure if there is any motivation to keep the node 8 support though. 🤔

ljharb:
Dropping support is something that should require motivation; retaining it should be the default. What's the concrete benefit of updating here, and/or dropping support, that would justify excluding those users?

package.json
```diff
@@ -26,7 +26,7 @@
   "dependencies": {
     "chalk": "^3.0.0",
     "cosmiconfig": "^6.0.0",
```
SimenB, line 28: if/when we upgrade, these should be bumped as well

mrtnzlml:
The motivation for this PR was to use the latest dependencies in our project and Node 8 didn't seem like a huge concern considering it's well beyond its end of life and the latest Jest and Eslint don't support such an old version either.

I do not have any other arguments for this PR - just the desire to be up to date.

PR #121, opened 2021-08-26
Upgrade to create-jest-runner 0.8.0

ZauberNerd:
This PR upgrades the dependency `create-jest-runner` to version `0.8.0` and also updates Jest to version `27`.

Further: We've added a clean-up commit for the `devDependencies`, because version ranges only make sense in `peerDependencies`.
If there is however a reason for the version ranges in the `devDependencies`, feel free to drop the commit :)

**Edit:** For some reason we get ESLint errors locally, although we didn't change the code and the eslint version shouldn't have changed significantly (https://github.com/jest-community/jest-runner-eslint/pull/121/files#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2deL2292)

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -21,11 +21,6 @@ jobs:
         node-version: [10.x, 12.x, 14.x, 16.x]
         eslint-version: [6, 7]
         jest-version: [25, 26, 27]
-        include:
-          # eslint@7 and jest@26 doesn't support node@8
-          - node-version: 8.x
-            eslint-version: 6
-            jest-version: 25
     runs-on: ubuntu-latest
 
     steps:
--- a/package.json
+++ b/package.json
@@ -26,13 +26,13 @@
   "dependencies": {
     "chalk": "^3.0.0",
     "cosmiconfig": "^6.0.0",
-    "create-jest-runner": "^0.6.0"
+    "create-jest-runner": "^0.8.0"
   },
   "devDependencies": {
     "@babel/cli": "^7.10.4",
     "@babel/core": "^7.10.4",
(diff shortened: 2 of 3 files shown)


ZauberNerd:
> In that case, there really aren't any options except:
> 
> 1. staying with the create-jest-runner we're using, for the duration
> 2. doing a breaking change and upgrading to the latest create-jest-runner - but this isn't sustainable over time; adding new jest versions should be semver-minor
> 3. create-jest-runner finds a way to support multiple jest versions such that we can use the latest version of it without a breaking change

2. How about releasing `jest-runner-eslint` (and also `create-jest-runner`) with the same major versions as the rest of the Jest packages?
3. The only viable option I see there, would be to make `jest-worker` a `peerDependency` of `create-jest-runner`.

/cc @SimenB do you see any other solutions?

ljharb:
Why force users to upgrade more things in sync??

SimenB:
I think we should drop Node 8 and Node 10 (but not any Jest or ESLint versions if we can). While I'm a fan of supporting older versions of node, at some point I don't think holding other upgrades back are worth it. And while Jest still supports node 10, I think we can just drop it while we're at it - running your linting on multiple versions of node isn't as useful as running your tests on multiple versions, so I don't think it's a huge loss of usability for this module (although running tests but not lint might require some config gymnastics, so maybe keeping 10 is a good idea for now).

ljharb:
The real solution would be if we could change create-jest-runner to be more backwards-compatible.

SimenB:
@ZauberNerd sorry about the long delay - available in https://github.com/jest-community/jest-runner-eslint/releases/tag/v2.0.0

PR #123, opened 2021-09-17
feat: support eslint v8

SimenB:
2 main breaking changes affect us (beyond node version requirement)

1. Removal of `CLIEngine`
2. Change of config overrides

The first one I believe to be no issue since we can check which is available and adjust our API usage accordingly (and it being async doesn't matter).

However, the second one stumped me a bit. It might be enough to - depending on the eslint version - use `overrideConfig` instead? This is the error:

```
Invalid Options:
- Unknown options: configFile, envs, fixDryRun, format, globals, ignorePattern, maxWarnings, parser, parserOptions, quiet, rules
- 'configFile' has been removed. Please use the 'overrideConfigFile' option instead.
- 'envs' has been removed. Please use the 'overrideConfig.env' option instead.
- 'globals' has been removed. Please use the 'overrideConfig.globals' option instead.
- 'ignorePattern' has been removed. Please use the 'overrideConfig.ignorePatterns' option instead.
- 'parser' has been removed. Please use the 'overrideConfig.parser' option instead.
- 'parserOptions' has been removed. Please use the 'overrideConfig.parserOptions' option instead.
- 'rules' has been removed. Please use the 'overrideConfig.rules' option instead.
- 'plugins' doesn't add plugins to configuration to load. Please use the 'overrideConfig.plugins' option instead.
- 'reportUnusedDisableDirectives' must be any of "error", "warn", "off", and null.
```

---

Note that I don't think we should necessarily merge this until ESLint v8 is released as stable, but it would be good to be ready to not block others.

---

Lastly, `eslint-plugin-import` explodes horribly due to using deep imports, so we should skip our lint run somehow, at least until that plugin adds support.


SimenB:
Hmm, one intermediary step could be to migrate to `ESLint` instead of `CLIEngine` when available - I assume that's something we can land without adding full v8 support, and it'll make it easier to land this

SimenB:
I've confirmed that commenting out the `no-unused-modules` import within the import plugin passes all tests with ESLint 8 🙂 So a release of that should allow us to land this (or at least unblock it of we want to wait)

SimenB:
🎉

ljharb:
I’d strongly suggest keeping this PR in draft until eslint v8 is fully released, to avoid needing to support prereleases.

SimenB:
Yup, agreed 🙂, just getting #128 is fine for me since it'll allow consumers to use v8 (albeit with peer dep warnings) instead of blocking them

PR #128, opened 2021-09-17
feat: use `ESLint` instead of `CLIEngine` when available

SimenB:
BREAKING CHANGE: Using `ESLint` might need different configuration
options. See https://eslint.org/docs/developer-guide/nodejs-api#parameters

`CLIEngine` is deprecated in ESLint v7 and removed entirely in ESLint v8, replaced by `ESLint`.

Landing this might give us support for ESLint v8 (#123) without changing anything more (although peer dep might still warn)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,6 +1,9 @@
 {
   "extends": ["airbnb-base", "prettier"],
   "plugins": ["prettier", "jest"],
+  "parserOptions": {
+    "ecmaVersion": 11
+  },
   "rules": {
     "no-underscore-dangle": 0,
     "arrow-body-style": 0,
--- a/integrationTests/__fixtures__/passing/jest-runner-eslint.config.js
+++ b/integrationTests/__fixtures__/passing/jest-runner-eslint.config.js
@@ -1,5 +1,8 @@
+const { ESLint } = require('eslint');
+
 module.exports = {
   cliOptions: {
-    global: ['hello'],
+    // `ESLint` requires this to be an object, not an array
+    global: ESLint ? { hello: true } : ['hello'],
   },
 };
--- a/package.json
+++ b/package.json
@@ -26,7 +26,8 @@
   "dependencies": {
     "chalk": "^3.0.0",
     "cosmiconfig": "^6.0.0",
-    "create-jest-runner": "^0.6.0"
+    "create-jest-runner": "^0.6.0",
+    "dot-prop": "^5.3.0"
   },
   "devDependencies": {
     "@babel/cli": "^7.10.4",
--- a/src/runner/__tests__/runESLint.test.js
+++ b/src/runner/__tests__/runESLint.test.js
@@ -11,7 +11,12 @@ const runESLintRunnerWithMockedEngine = options => {
 
       executeOnFiles() {
         return {
+          results:
+            options.cliEngine.errorCount > 0
+              ? [{ errorCount: options.cliEngine.errorCount, warningCount: 0 }]
+              : [],
           errorCount: options.cliEngine.errorCount,
+          warningCount: 0,
         };
       }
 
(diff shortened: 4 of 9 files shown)


integrationTests/__fixtures__/passing/jest-runner-eslint.config.js
```diff
@@ -1,5 +1,8 @@
+const { ESLint } = require('eslint');
+
 module.exports = {
   cliOptions: {
-    global: ['hello'],
+    // `ESLint` requires this to be an object, not an array
+    global: ESLint ? { hello: true } : ['hello'],
```
SimenB, line 6: this is main reason I think this is a breaking change

.eslintrc.json
```diff
@@ -1,6 +1,9 @@
 {
   "extends": ["airbnb-base", "prettier"],
   "plugins": ["prettier", "jest"],
+  "parserOptions": {
+    "ecmaVersion": 11
```
SimenB, line 5: needed for optional chaining (babel transpiles it away)

SimenB: thinking about it, #125 _might_ be considered breaking as well, if people previously ignored an engine warning and used this package on node 6. I don't consider it a breaking change, but bundling these together in a new major removes any ambiguity 🤷

integrationTests/__fixtures__/passing/jest-runner-eslint.config.js
SimenB: thinking about it, #125 _might_ be considered breaking as well, if people previously ignored an engine warning and used this package on node 6. I don't consider it a breaking change, but bundling these together in a new major removes any ambiguity 🤷

ljharb: It seems pretty valuable to try to support both eslint 8 and < 8 at the same time

package.json
```diff
@@ -26,7 +26,8 @@
   "dependencies": {
     "chalk": "^3.0.0",
     "cosmiconfig": "^6.0.0",
-    "create-jest-runner": "^0.6.0"
+    "create-jest-runner": "^0.6.0",
+    "dot-prop": "^5.3.0"
```
ljharb, line 30: This author is planning to make all their packages ESM-only - I’d strongly suggest avoiding adding them as a dependency.

integrationTests/__fixtures__/passing/jest-runner-eslint.config.js
SimenB: sure, but I don't think it's our responsibility to coerce user input into the correct type. Consumers should configure the the version of ESLint they choose to use correctly

package.json
SimenB: I went for v5 since v6 doesn't support node 8. We can migrate to some other module later if needed. Unless you have a good suggestion for a module that's not this or lodash? 😀

integrationTests/__fixtures__/passing/jest-runner-eslint.config.js
ljharb: Oh true, fair point - i think it’s 100% fine if a user switches to eslint 8, and this runner breaks until they also fix their configuration (altho hopefully they get a very clear error message)

SimenB: Yeah, the error message from eslint is pretty good 👍  Specifically in this case if I leave it as an array

```
ESLint configuration in CLIOptions is invalid:
      - Property "globals" is the wrong type (expected object but got `["hello"]`).
```

package.json
SimenB: not sure - the `ESLint` class requires a bunch of stuff in `configOverrides` instead of in the root object, and deep merging is quite hellish in JS (without some module). Going with a dotted setter seemed like a nice shortcut for the specific CLI options we've added support for

(I'm not entirely sure why we have this normalization instead of passing through raw as-is, but that's a separate discussion)

SimenB: I'll land this, and if we figure out some cleaner way of dealing with this we can revisit. It's not exposed, so it's an implementation detail we can change at will

PR #134, opened 2022-03-31
docs: correct link to "projects" Jest documentation

ivangermanov:
### Current problem
Link to `projects` section on Jest docs doesn't work.
### Fix
Fix the link to the correct URL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ Please update `testMatch` to match your project folder structure
 
 #### Alongside other runners
 
-It is recommended to use the [`projects`](https://facebook.github.io/jest/docs/en/configuration.html#projects-array-string-projectconfig) configuration option to run multiple Jest runners simultaneously.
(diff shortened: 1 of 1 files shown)


SimenB (approved):
thanks!

ljharb:
Links should never die (it’s a core principle of the web)  so has an issue been filed on jest to redirect the old link to the new one?

SimenB:
It's just the anchor, it still goes to the correct site

ljharb:
That’s good, but old anchors can be preserved with `<a id=“blah”></a>`, and that should still be done.

SimenB:
docs are markdown, we don't write the ids manually. Maybe Docusaurus should be updated to generate both old and new style anchors, but I don't think it's a biggie in this case

ljharb:
Not the current IDs; but when renaming a section, you can add that HTML to markdown to preserve the old one - it works just fine.

SimenB:
Right, my point is that we didn't rename/change anything in the docs themselves, just changed the version of the site generator

ljharb:
Oh wow, docusaurus broke the link anchors on its own? That’s pretty terrifying.

PR #135, opened 2022-05-02
Allow Jest v28

lencioni:
Release post:

https://jestjs.io/blog/2022/04/25/jest-28


ljharb (approved):

package.json
```diff
@@ -42,16 +42,16 @@
     "eslint-plugin-jest": "^23.17.1",
     "eslint-plugin-prettier": "^3.1.4",
     "execa": "^3.4.0",
-    "jest": "^25.1 || ^26 || ^27",
+    "jest": "^25.1 || ^26 || ^27 || ^28",
     "jest-watch-select-projects": "^2.0.0",
-    "jest-watch-typeahead": "^0.6.4",
+    "jest-watch-typeahead": "^1.1.0",
```
SimenB, line 47: this only works in jest 27 and newer (as v1 is native ESM), but I guess watch mode isn't the most important thing

ljharb: Does v0.x support jest 28? If so, we should probably downgrade it.

SimenB: yes, the interface of watch plugins is unchanged

ljharb: ah, then it’s still invalid - no chance of a v0.x bump with updated peer deps?

SimenB: we could upgrade to yarn v2 and use [`packageExtension`](https://yarnpkg.com/configuration/yarnrc#packageExtensions). Either way, it's just a warning locally, it won't be visible to consumers

lencioni: What do you think about changing the semver here to be `^0.6.4 || ^1.1.0`?

SimenB: that works 🙂

lencioni: https://github.com/jest-community/jest-runner-eslint/pull/136

ljharb: @SimenB yarn-specific features aren’t useful. #136 works tho

SimenB: Since it only applies to this specific repo and no consumers, and we use yarn, I think it would work fine. but changing the range works as well

PR #140, opened 2022-05-22
Update chalk to version 4.0.0

jsfix-updater:


ljharb (changes requested):
Why? this package should be removed rather than updated, since v5 is ESM-only, and thus we can’t ever get the latest version anyways.

PR #146, opened 2022-12-01
feat: Split results to work with github-actions reporter

cprussin:
This PR fixes the reporting to report separate messages for each eslint error, so that errors can be reported with the [github actions reporter](https://github.com/facebook/jest/blob/main/packages/jest-reporters/src/GitHubActionsReporter.ts).  Frankly -- it's a bit hacky because the github actions reporter relies on stack frames to figure out location information, rather than using the `location` field of the message, so I'm adding a fake stack line to the message.

It doesn't change the output significantly for other reporters, apart from changing `✕ ESLint` to actually report the list of failed lint tests, e.g. `✕ no-undef\n✕ no-undef`.

This PR also makes the integration test output normalization code a bit more restricted, because the prior implementation was normalizing things that don't need to be normalized (i.e. relative paths in stack frames) and was eating the output of the github action reporting, making it impossible to actually test the action reporter.

**EDIT:** I also added a commit to upgrade `create-jest-runner` since I realized the version we were previously on was creating a `failureMessages` array as `[undefined]` instead of `[]` when calling `pass`, which caused the github actions reporter to report an empty failure message on every single file.


cprussin:
@SimenB I believe you're the one to ask for a review on this, please!  Thanks a bunch!

package.json
```diff
@@ -26,7 +26,7 @@
   "dependencies": {
     "chalk": "^3.0.0",
     "cosmiconfig": "^6.0.0",
-    "create-jest-runner": "^0.6.0",
+    "create-jest-runner": "^0.12.2",
```
SimenB, line 29: this is a breaking change, so we cannot do it (or, it should at least be in a separate PR)

src/runner/runESLint.js
```diff
@@ -97,11 +98,51 @@ const runESLint = async ({ testPath, config, extraOptions }) => {
...
 
   if (report[0]?.errorCount > 0) {
-    return fail({
-      start,
-      end,
-      test: { path: testPath, title: 'ESLint', errorMessage: message },
-    });
+    const startTime = new Date(start).getTime();
+    const endTime = new Date(end).getTime();
+    return {
```
SimenB, line 103: having to manually construct the entire result is odd - we should still be able to use `fail` to report many errors, no?

SimenB (changes requested):
exciting stuff!

src/runner/runESLint.js
cprussin: Yeah, it's a shame, but [`fail` does not support multiple errors](https://github.com/jest-community/create-jest-runner/blob/main/lib/fail.ts#L46-L48).

[`toTestResults`](https://github.com/jest-community/create-jest-runner/blob/main/lib/toTestResult.ts#L65) actually [does](https://github.com/jest-community/create-jest-runner/blob/main/lib/toTestResult.ts#L50) but `toTestResults` isn't exposed through the main module.

We have a few options:

1. I can put in a PR to `create-jest-runner` to expose `toTestResults` correctly.
2. I can put in a PR to `create-jest-runner` to allow multiple failures in `fail`.
3. We can do an internal require to `create-jest-runner/toTestResults` instead.

My preference would be either to do #3 or to move forward with this PR as is.  I'll follow up by doing #1 AND #2 on `create-jest-runner` and whenever that lands, I'll come back here and put in another PR to bump the version and use the new shininess.

Thoughts?

package.json
SimenB: Still separate, yeah - fixing output bug is not the same as adding more precise errors, and it's easier to look at them separately. But that PR can manually filter the results, perhaps? Instead of bumping the dep

src/runner/runESLint.js
SimenB: We should do both 1 and 2, yes 👍 I don't think we need 3 - doing it manually like you do here is fine in that case. Perhaps leave a code comment saying so?

package.json
cprussin: Well, we can't filter the results in this codebase anywhere if we're using the `pass` function, since the bug is between that function and the underlying jest code (actually the bug is in [`toTestResult`](https://github.com/jest-community/create-jest-runner/blob/v0.6.0/lib/toTestResult.js#L35) in v0.6.0).  But what we could do is to replace the `pass` call with manually constructing the result like I'm doing above, and that would prevent having to bump the dependency.  Is that your preference?

cprussin: OK, I've added a commit to not upgrade `create-jest-runner` and to construct the results manually for `pass` (and `skip`, which I had missed before).  Also added a comment above the function explaining why we're doing it.  Let me know if this approach works now.

src/runner/runESLint.js
cprussin: Added a comment to that effect in the latest commit

cprussin:
hey @SimenB , anything else I can do to get this PR good to merge?  Thanks a bunch!

src/runner/runESLint.js
```diff
@@ -1,7 +1,94 @@
...
+ *
+ * 2. The `fail` function in `create-jest-runner` does not support passing
+ *    multiple failure messages, which makes it impossible to annotate each
+ *    eslint failure.  This has not been resolved, although presumably could be
+ *    worked around by using the underlying `toTestResult` function instead
+ *    (although that function isn't exposed in the public API of that library).
+ *
+ * TODO At some point, we should put a PR in to `create-jest-runner` to resolve
+ * point 2 above, and then should upgrade and remove this function and go back
+ * to using `pass`, `skip`, and `fail` from that library instead.
```
SimenB, line 23: are you up for sending that PR?

SimenB (approved):
thanks!

SimenB (changes requested):
Ergh, CI is unhappy - we need to filter out the times from the snapshots

src/runner/runESLint.js
cprussin: Yeah, I can do that.  I just realized you own both of these repos -- do you have a preference for me to do that before / after this lands?

cprussin:
@SimenB Looks like the vast majority of the failures were caused by me not accounting for jest <v28, where there is no github action runner, or eslint 6/7/8 having slightly different stack messages for the custom formatter error (those stacks were previously chomped by the normalization but no longer are).  I've fixed both cases in https://github.com/jest-community/jest-runner-eslint/pull/146/commits/40d19b3344d9630274246588558e44f123882cf9.

It's weird that you saw some cases failing due to the times as the regex I have (and the snapshots) _are_ filtering out times, so something must be weird with some specific combination of versions.  In any case, I didn't find that exact failure looking through the build matrix, but it was a lot to look through so it probably got lost in all the other failures.  Would you mind running the actions again with my latest commit so I can more easily filter down to that failure and then I can fix it?

Thanks!

src/runner/runESLint.js
cprussin: also, can you advise on if you want `toTestResult` to get exposed vs. just adding support for multiple pass/fail/skip/todo messages?

SimenB: > do you have a preference for me to do that before / after this lands?

No preference on timing - this repo is stuck on an older version for now anyways.

> also, can you advise on if you want `toTestResult` to get exposed vs. just adding support for multiple pass/fail/skip/todo messages?

happy to expose the underlying helper as well

cprussin:
@SimenB dang, I forgot to add an empty function argument to my skip call.  I'll fix that but there's still likely other issues on different version combos -- any chance you can permission me to run the workflows myself so I don't have to keep bugging you while I work through the old version issues?

SimenB:
If you send a PR for something else trivial that I can land, then GH action should run automatically

cprussin:
@SimenB  I'm fairly sure I've fixed the issues now.  If you'd like to give me access to run the checks myself in case I'm wrong, I put in https://github.com/jest-community/jest-runner-eslint/pull/149.  But no need if you're willing to just keep kicking off the run for me, I feel good this time around :)

Sorry for all the back and forth here and thanks for the patience.

SimenB:
Cool, merged 👍

cprussin:
@SimenB as always, regexes are a pain, but I got them nailed down.  looks like I got all the tests passing :tada: .  Good to go now or any other tweaks you'd like?

src/runner/runESLint.js
```diff
@@ -1,7 +1,94 @@
...
+      testFilePath: testPath,
+      numPassingAsserts: numPassingTests,
+      status: result.status,
+      title: result.title,
+    })),
+  };
+};
+
+const mkAssertionResults = (testPath, report) =>
+  report[0].messages?.map(reportMessage => ({
```
SimenB, line 78: why only `report[0]`? Is there never more than one?

SimenB (approved):
thanks!

src/runner/runESLint.js
cprussin: That's a good question, there's only ever one, but only because each result represents a file and on [line 186](https://github.com/jest-community/jest-runner-eslint/blob/aa2856c4d4715cfe0727fa63f97484dd32ae6f18/src/runner/runESLint.js#L186) we are passing in just a single file at a time to the eslint api.

I suppose I could change this so that it flat maps across the messages inside arbitrary number of reports just in case we were to ever move to using eslint in batch mode for whatever reason, but there's a lot more that wasn't added in this PR that would have to be changed too, such as [line 198](https://github.com/jest-community/jest-runner-eslint/blob/aa2856c4d4715cfe0727fa63f97484dd32ae6f18/src/runner/runESLint.js#L198), [line 212](https://github.com/jest-community/jest-runner-eslint/blob/aa2856c4d4715cfe0727fa63f97484dd32ae6f18/src/runner/runESLint.js#L212), and [line 239](https://github.com/jest-community/jest-runner-eslint/blob/aa2856c4d4715cfe0727fa63f97484dd32ae6f18/src/runner/runESLint.js#L239), all of which are making the same assumption.  So when I was writing this, it definitely felt icky, but it didn't feel like the right PR to make that change in.

Happy to make that change if you'd prefer though, it's pretty simple.

SimenB: Sounds like a separate PR. Wanna send that one? 🙂

SimenB:
@cprussin sorry about the delay - https://github.com/jest-community/jest-runner-eslint/releases/tag/v1.2.0 is out

src/runner/runESLint.js
SimenB: @cprussin came over this again now - would still love to see that PR for `fail` 😀

cprussin: Hmm jog my memory, I left the following comment:

> This has been resolved in later versions of `create-jest-runner`, but upgrading is a breaking change that is incompatible with some node versions we support.

Is that still accurate?  If so, even if I put that PR in for `fail`, it won't change anything here since we still can't upgrade right?

SimenB: That is correct 👍

PR #176, opened 2023-04-27
Support eslint's new flat config

### LINKED ISSUE #166: Support flat config
As you all probably already know, [there is a new eslint config format in the works](https://eslint.org/blog/2022/08/new-config-system-part-1/).

It seems like in order for libraries to work with it, the libraries [must use the `FlatESLint` class instead of the `ESLint` class](https://eslint.org/blog/2022/08/new-config-system-part-3/#using-flat-config-with-the-eslint-class), at least tentatively while the new config system is being stabilized and everything is getting migrated.

Would there be any interest in supporting the new flat config format during the transitional phase?  Presumably it would be a pretty simple change -- just would need to do something like add a setting that switches between `ESLint` and `FlatESLint` [here](https://github.com/jest-community/jest-runner-eslint/blob/main/src/runner/runESLint.js#L119).  But I could also see an argument against supporting an experimental API at all, especially given the article above mentions that the `FlatESLint` class would only exist during the transition phase and would go away once flat config becomes the default.

I'd be happy to submit a PR if the maintainers here would be amenable to accepting it.

@SimenB I'll tag you to get your opinion on this since you helped me merge in my last contributions and I know you know the ecosystem really well, but I know you're pretty busy so please help me tag in someone else who can give guidance / opinions if there's someone more appropriate!

Thanks as always

**EDIT:** I hacked up a super quick implementation of what a switch to `FlatESLint` would look like [here](https://github.com/jest-community/jest-runner-eslint/compare/main...cprussin:jest-runner-eslint:main) which I've tested in my own configs, if you want to take a look.  This implementation doesn't add an option, it just switches over, and it doesn't do anything to try to translate changed cli options to proper equivalent (I'm just deleting those options entirely).  So this wouldn't be ready to be put into a PR, but it gives you the gist of the changes that would need to happen.

cprussin:
Fixes #166 .

Note I also had to change the strategy for running integration tests is this PR.  Prior to this PR, integration tests ran in a child process where the cwd was the root of the integration tests, and jest projects were used to select individual integration test directories.  However, [eslint will look for the flat config starting in the process cwd](https://github.com/eslint/eslint/blob/d85efad655deacc0dc3fdbbace33307094c3b91b/lib/cli.js#L299), so with that strategy the flat config would have to be placed in a common path to all integration tests, and thus the integration test runs would always detect that they're running in flat config mode.

As of this PR, the integration tests run with a child process where the cwd is the individual integration test fixture path as well.  Practically, this required updating some snapshots and moving some `.eslintignore` entries into the test fixture folders.

### DIFF
--- a/integrationTests/flat-config.test.js
+++ b/integrationTests/flat-config.test.js
@@ -1,0 +1,10 @@
+const { version } = require('eslint/package.json');
+const semver = require('semver');
+const runJest = require('./runJest');
+
+(semver.satisfies(version, '>=8.41') ? it : it.skip)(
+  'Works with the new flat config format',
+  async () => {
+    expect(await runJest('flat-config')).toMatchSnapshot();
+  },
+);
--- a/src/runner/runESLint.js
+++ b/src/runner/runESLint.js
@@ -1,6 +1,32 @@
 const { ESLint } = require('eslint');
 const getESLintOptions = require('../utils/getESLintOptions');
 
+let FlatESLint;
+let shouldUseFlatConfig;
+
+try {
+  // Use a dynamic require here rather than a global require because this
+  // import path does not exist in eslint v7 which this library still
+  // supports
+  //
+  // ESlint exposes the new FlatESLint API under `eslint/use-at-your-own-risk` by
+  // using it's [export configuration](https://tinyurl.com/2s45zh9b).  However,
+  // the `import/no-unresolved` rule is [not aware of
+  // `exports`](https://tinyurl.com/469djpx3) and causes a false error here.  So,
+  // let's ignore that rule for this import.
+  //
+  // eslint-disable-next-line global-require, import/no-unresolved
+  const eslintExperimental = require('eslint/use-at-your-own-risk');
+  FlatESLint = eslintExperimental.FlatESLint;
+  shouldUseFlatConfig = eslintExperimental.shouldUseFlatConfig;
+} catch {
+  /* no-op */
+}
+
+if (shouldUseFlatConfig === undefined) {
+  shouldUseFlatConfig = () => Promise.resolve(false);
+}
+
 /*
  * This function exists because there are issues with the `pass`, `skip`, and
  * `fail` functions from `create-jest-runner`:
@@ -96,8 +122,41 @@ const getComputedFixValue = ({ fix, quiet, fixDryRun }) => {
   return undefined;
 };
 
+const getESLintConstructor = async () => {
+  if (await shouldUseFlatConfig()) {
+    return FlatESLint;
+  }
+
+  return ESLint;
+};
+
+// Remove options that are not constructor args.
+const getESLintConstructorArgs = async cliOptions => {
+  // these are not constructor args for either the legacy or the flat ESLint
+  // api
+  const { fixDryRun, format, maxWarnings, quiet, ...legacyConstructorArgs } =
+    cliOptions;
+
+  if (await shouldUseFlatConfig()) {
+    // these options are supported by the legacy ESLint api but aren't
+    // supported by the ESLintFlat api
+    const {
+      extensions,
+      ignorePath,
+      rulePaths,
+      resolvePluginsRelativeTo,
+      useEslintrc,
+      overrideConfig,
+      ...flatConstructorArgs
+    } = legacyConstructorArgs;
+    return flatConstructorArgs;
+  }
+
+  return legacyConstructorArgs;
+};
+
 let cachedValues;
-const getCachedValues = (config, extraOptions) => {
+const getCachedValues = async (config, extraOptions) => {
   if (!cachedValues) {
     const { cliOptions: baseCliOptions } = getESLintOptions(config);
     const cliOptions = {
@@ -106,20 +165,20 @@ const getCachedValues = (config, extraOptions) => {
       ...extraOptions,
     };
 
-    // these are not constructor args, so remove them
-    const { fixDryRun, format, maxWarnings, quiet, ...eslintOptions } =
-      cliOptions;
-
-    const cli = new ESLint(eslintOptions);
+    const ESLintConstructor = await getESLintConstructor();
+    const cli = new ESLintConstructor(
+      await getESLintConstructorArgs(cliOptions),
+    );
 
     cachedValues = {
       isPathIgnored: cli.isPathIgnored.bind(cli),
       lintFiles: (...args) => cli.lintFiles(...args),
       formatter: async (...args) => {
-        const formatter = await cli.loadFormatter(format);
+        const formatter = await cli.loadFormatter(cliOptions.format);
         return formatter.format(...args);
       },
       cliOptions,
+      ESLintConstructor,
     };
   }
 
@@ -139,10 +198,8 @@ const runESLint = async ({ testPath, config, extraOptions }) => {
     config.setupFilesAfterEnv.forEach(require);
   }
 
-  const { isPathIgnored, lintFiles, formatter, cliOptions } = getCachedValues(
-    config,
-    extraOptions,
-  );
+  const { isPathIgnored, lintFiles, formatter, cliOptions, ESLintConstructor } =
+    await getCachedValues(config, extraOptions);
 
   if (await isPathIgnored(testPath)) {
     return mkTestResults({
@@ -163,13 +220,13 @@ const runESLint = async ({ testPath, config, extraOptions }) => {
   const report = await lintFiles([testPath]);
 
   if (cliOptions.fix && !cliOptions.fixDryRun) {
-    await ESLint.outputFixes(report);
+    await ESLintConstructor.outputFixes(report);
   }
 
   const end = Date.now();
 
   const message = await formatter(
-    cliOptions.quiet ? ESLint.getErrorResults(report) : report,
+    cliOptions.quiet ? ESLintConstructor.getErrorResults(report) : report,
   );
 
   if (report[0]?.errorCount > 0) {
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -19,7 +19,7 @@ jobs:
       fail-fast: false
       matrix:
         node-version: [12.x, 14.x, 16.x, 18.x, 19.x]
-        eslint-version: [7, 8]
+        eslint-version: [7, "8.40", 8]
         jest-version: [27, 28, 29]
         jest-watch-typeahead-version: [1, 2]
         exclude:
--- a/integrationTests/__fixtures__/flat-config/__eslint__/file.js
+++ b/integrationTests/__fixtures__/flat-config/__eslint__/file.js
@@ -1,0 +1,3 @@
+const a = 1;
+
+console.log('a', a);
--- a/integrationTests/__fixtures__/flat-config/eslint.config.js
+++ b/integrationTests/__fixtures__/flat-config/eslint.config.js
@@ -1,0 +1,7 @@
+module.exports = [
+  {
+    rules: {
+      'no-console': 'error',
+    },
+  },
+];
--- a/integrationTests/__fixtures__/flat-config/jest.config.js
+++ b/integrationTests/__fixtures__/flat-config/jest.config.js
@@ -1,0 +1,4 @@
+module.exports = {
+  runner: '../../../',
+  testMatch: ['**/__eslint__/**/*.js'],
+};
--- a/integrationTests/__fixtures__/legacy-config-at-eslint.config.js/__eslint__/file.js
+++ b/integrationTests/__fixtures__/legacy-config-at-eslint.config.js/__eslint__/file.js
@@ -1,0 +1,3 @@
+const a = 1;
+
+console.log('a', a);
--- a/integrationTests/__fixtures__/legacy-config-at-eslint.config.js/eslint.config.js
+++ b/integrationTests/__fixtures__/legacy-config-at-eslint.config.js/eslint.config.js
@@ -1,0 +1,5 @@
+module.exports = {
+  rules: {
+    'no-console': 'error',
+  },
+};
--- a/integrationTests/__fixtures__/legacy-config-at-eslint.config.js/jest-runner-eslint.config.js
+++ b/integrationTests/__fixtures__/legacy-config-at-eslint.config.js/jest-runner-eslint.config.js
@@ -1,0 +1,3 @@
+module.exports = {
+  config: './eslint.config.js',
+};
--- a/integrationTests/__fixtures__/legacy-config-at-eslint.config.js/jest.config.js
+++ b/integrationTests/__fixtures__/legacy-config-at-eslint.config.js/jest.config.js
@@ -1,0 +1,4 @@
+module.exports = {
+  runner: '../../../',
+  testMatch: ['**/__eslint__/**/*.js'],
+};
--- a/integrationTests/__snapshots__/flat-config.test.js.snap
+++ b/integrationTests/__snapshots__/flat-config.test.js.snap
@@ -1,0 +1,19 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Works with the new flat config format 1`] = `
+"FAIL __eslint__/file.js
+  ✕ no-console 
+
+
+/mocked-path-to-jest-runner-mocha/integrationTests/__fixtures__/flat-config/__eslint__/file.js
+  3:1  error  Unexpected console statement  no-console
+
+✖ 1 problem (1 error, 0 warnings)
+
+Test Suites: 1 failed, 1 total
+Tests:       1 failed, 1 total
+Snapshots:   0 total
+Time:        
+Ran all test suites.
+"
+`;
--- a/integrationTests/__snapshots__/legacy-config-at-eslint.config.js.test.js.snap
+++ b/integrationTests/__snapshots__/legacy-config-at-eslint.config.js.test.js.snap
@@ -1,0 +1,22 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Does not try to use flat config on eslint versions that don't support it 1`] = `
+"PASS __eslint__/file.js
+  ● Console
+
+  console.warn
+    
+    /mocked-path-to-jest-runner-mocha/integrationTests/__fixtures__/legacy-config-at-eslint.config.js/__eslint__/file.js
+      3:1  warning  Unexpected console statement  no-console
+    
+    ✖ 1 problem (0 errors, 1 warning)
+
+  ✓ ESLint 
+
+Test Suites: 1 passed, 1 total
+Tests:       1 passed, 1 total
+Snapshots:   0 total
+Time:        
+Ran all test suites.
+"
+`;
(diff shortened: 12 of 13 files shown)


cprussin:
cc @SimenB @ljharb

ljharb:
This seems worth splitting into two PRs - one changing the way the tests work, and another after that lands that adds support for eslint's flat config?

package.json
```diff
@@ -27,7 +27,8 @@
     "chalk": "^4.0.0",
     "cosmiconfig": "^7.0.0",
     "create-jest-runner": "^0.11.2",
-    "dot-prop": "^6.0.1"
+    "dot-prop": "^6.0.1",
+    "find-up": "^5.0.0"
```
ljharb, line 31: let's please not add packages from this maintainer, since they're converting to ESM-only and make frequent breaking changes.

ljharb (changes requested):

cprussin:
I could @ljharb but it seems like unnecessary churn, especially if the two PRs will land one after another and if the only motivation to do the one is in order to introduce the other.

package.json
cprussin: I only added this because it's [what eslint uses](https://github.com/eslint/eslint/blob/d85efad655deacc0dc3fdbbace33307094c3b91b/lib/eslint/flat-eslint.js#L266-L271) and I think we should try to match the functionality as much as possible.  Ideally the `findFlatConfigFile` function (or even `shouldUseFlatConfig`](https://github.com/eslint/eslint/blob/d85efad655deacc0dc3fdbbace33307094c3b91b/lib/cli.js#L283-L301) would be exposed directly by eslint but it isn't, so we need to mimic it.

ljharb:
The point would be to a) reduce the git diff size, and b) prove that the test refactors pass without changing the production code, both of which reduce the risk of any change.

package.json
ljharb: Fair; pinning to v5 is fine i guess for now.

Regarding what eslint exposes, if we have a need for it, please file an issue on eslint, and they may be willing to expose it :-)

cprussin:
> The point would be to a) reduce the git diff size, and b) prove that the test refactors pass without changing the production code, both of which reduce the risk of any change.

Fair enough, I'll split it up tomorrow and will investigate the failing tests tomorrow.  New PRs inbound soon.

package.json
cprussin: Yep that's fair -- I'll put in an issue / PR to eslint and see if they're willing to do so.  Will keep this in this PR for now but if things move forward on that side, I'll either update this PR or put in a follow up.

ljharb:
Let's try to send just one new PR, and update this one to be one of the two parts :-) Thanks!

package.json
cprussin: In case you're following along, eslint issue is here: https://github.com/eslint/eslint/issues/17129

I'll submit a PR over there if/when the eslint folks get back to me on which strategy would be acceptable.

cprussin:
@ljharb I'm going to put this on hold for a bit for two reasons:

1. Pending the reply in https://github.com/eslint/eslint/issues/17129, that could significantly simplify things -- ideally the eslint folks will support exposing an API that switches between `ESLint` and `FlatESLint` automatically and that would cut down on much of the change in this PR.

2. I'm also fairly sure the `CLIEngine` code is not necessary any more as `jest-runner-eslint` no longer supports older versions of eslint where that would have been applicable.  I'll double check that and will submit a separate PR to strip that out, which will further simplify this.

If both of those last 2 work out, and once https://github.com/jest-community/jest-runner-eslint/pull/177 gets merged, this PR will turn into effectively just a one-line change of an import path

I'll keep the PR open until I wrap those other changes up and then will rebase accordingly and ping you when I think it's ready to merge.

cprussin:
@ljharb I think this PR is good to merge now if you'd like.  I'm guessing it will take the eslint folks some time to get back to me on https://github.com/eslint/eslint/issues/17129 and then I'll need to implement and PR and merge whenever they suggest.  I'm happy to wait on that and then update this PR once something lands, or also happy to get this merged in and then go back and put in a follow-up PR whenever the eslint side moves forward.

integrationTests/flat-config.test.js
```diff
@@ -0,0 +1,14 @@
+const { version } = require('eslint/package.json');
+const semver = require('semver');
+const runJest = require('./runJest');
+
+// ESLint <8 doesn't support the flat config format at all, so this test would
+// always fail.  But we need the skipped test so we don't get a failure for `1
+// snapshot obsolete`.
+if (semver.satisfies(version, '7')) {
```
ljharb, line 8:
```suggestion
if (semver.satisfies(version, '< 8')) {
```
so it matches the intention.

integrationTests/flat-config.test.js
```diff
@@ -0,0 +1,14 @@
...
+  it.skip('Works with the new flat config format', () => {});
+} else {
+  it('Works with the new flat config format', async () => {
```
ljharb, line 11:
```suggestion
(semver.satisfies(version, '>= 8') ? it : it.skip)('Works with the new flat config format', async () => {
```
another alternative that doesn't need a redundant comment

src/runner/runESLint.js
```diff
@@ -96,8 +98,55 @@ const getComputedFixValue = ({ fix, quiet, fixDryRun }) => {
   return undefined;
 };
 
+const useFlatConfig = async () =>
+  !!(await findUp('eslint.config.js', { cwd: process.cwd() }));
```
ljharb, line 102: tbh i'm not sure this is sufficient, since `eslint -c eslint.config.js` would be the legacy config system in eslint 7.

src/runner/runESLint.js
```diff
@@ -96,8 +98,55 @@ const getComputedFixValue = ({ fix, quiet, fixDryRun }) => {
...
+
+  return ESLint;
+};
+
+// Remove options that are not constructor args.
+const getESLintConstructorArgs = async cliOptions => {
+  // these are not constructor args for either the legacy or the flat ESLint
+  // api
+  const { fixDryRun, format, maxWarnings, quiet, ...legacyConstructorArgs } =
+    cliOptions;
```
ljharb, line 128:
```suggestion
  const {
    fixDryRun,
    format,
    maxWarnings,
    quiet,
    ...legacyConstructorArgs
  } = cliOptions;
```

integrationTests/flat-config.test.js
cprussin: I find that to be obtuse and hard to quickly scan, would you object to just extracting out the comment as a constant?

src/runner/runESLint.js
cprussin: hmm I think this one was done by prettier automatically....

integrationTests/flat-config.test.js
ljharb: i'm not sure what you mean as "the comment" but yes, a constant for `isESLint8` is fine.

src/runner/runESLint.js
ljharb: yeah prettier often makes code uglier, but i hope it wouldn't reformat my suggestion?

integrationTests/flat-config.test.js
cprussin: Sorry, that was poor phrasing, what I meant was this:

```js
const TEST_NAME = 'Works with the new flat config format';

if (semver.satisfies(version, '>= 8')) {
    it(TEST_NAME, async () => ...);
} else {
    it.skip(TEST_NAME, () => {});
}

File: src/runner/runESLint.js (hunk 5 above)
Comment (cprussin) [reply src/runner/runESLint.js:128]: I actually had it as you have it originally and this is what prettier reformatted it to

Comment (ljharb) [reply src/runner/runESLint.js:128]: gross, ok, guess we'll have to live with the uglier code :-/

File: src/runner/runESLint.js (hunk 4 above)
Comment (cprussin) [reply src/runner/runESLint.js:102]: [I'm not sure if that's true](https://github.com/eslint/eslint/blob/559ff4e4bc54a8b6e6b54825d83c532d724204b3/lib/options.js#L153-L160) because reading through the option code it looks like `-c` only affects the name of the config file, regardless of the config system.  It looks like the flat config is selected in one of two cases:

1. [`eslint.config.js` exists in the tree](https://github.com/eslint/eslint/blob/559ff4e4bc54a8b6e6b54825d83c532d724204b3/lib/eslint/flat-eslint.js#L266-L271)
2. [`process.env.SHOULD_USE_ESLINT` is 'true'](https://github.com/eslint/eslint/blob/559ff4e4bc54a8b6e6b54825d83c532d724204b3/lib/cli.js#L283-L301)

I could add the case for the `SHOULD_USE_ESLINT` env var but I don't see in the eslint code where `-c` would use the old config format.  I'm no expert though so please correct me if I'm mistaken.

File: integrationTests/flat-config.test.js (hunk 3 above)
Comment (cprussin) [reply integrationTests/flat-config.test.js:11]: eh, fair enough, I don't really care much on this.  I'll change as you suggested.

File: src/runner/runESLint.js (hunk 4 above)
Comment (ljharb) [reply src/runner/runESLint.js:102]: I meant in eslint 7, which we still support.

Comment (cprussin) [reply src/runner/runESLint.js:102]: Ahh right, I can add a semver check in there too, do you think that's the right fix?

Comment (ljharb) [reply src/runner/runESLint.js:102]: That seems reasonable, assuming 8.0.0 is what has flat config in it.

Comment (cprussin) [reply src/runner/runESLint.js]: fixed, thanks

**Comment by cprussin** [comment]
@ljharb  OK I think that should do it, I added an integration test to be sure we don't try to load the config using the flat format on eslint versions that don't support it.  Let me know if there's anything else you'd suggest.

Thanks for the reviews & help getting these merged!

**Review by ljharb — APPROVED** [review]
this should be rebased, but otherwise lgtm. not going to merge without other reviews tho

**Comment by cprussin** [comment]
rebased, thanks @ljharb

File: src/runner/runESLint.js [hunk 6]
```diff
@@ -1,4 +1,8 @@
 const { ESLint } = require('eslint');
+const { version } = require('eslint/package.json');
+const semver = require('semver');
+const findUp = require('find-up');
```
SimenB, line 4: Could we sort these?

src/runner/runESLint.js
```diff
@@ -96,8 +100,65 @@ const getComputedFixValue = ({ fix, quiet, fixDryRun }) => {
   return undefined;
 };
 
+const useFlatConfig = async () => {
+  if (semver.satisfies(version, '>=8')) {
```
SimenB, line 104: Extract out of function?

src/runner/runESLint.js
```diff
@@ -96,8 +100,65 @@ const getComputedFixValue = ({ fix, quiet, fixDryRun }) => {
...
+    switch (process.env.ESLINT_USE_FLAT_CONFIG) {
+      case 'true':
+        return true;
+      case 'false':
+        return false;
+      default:
+        return !!(await findUp('eslint.config.js', { cwd: process.cwd() }));
```
SimenB, line 111: IMO this is hard to read. Could you create a scope within the `default` case and extract `findUp` call to a variable

src/runner/runESLint.js
```diff
@@ -96,8 +100,65 @@ const getComputedFixValue = ({ fix, quiet, fixDryRun }) => {
...
+  }
+  return false;
+};
+const getESLintConstructor = async () => {
+  if (await useFlatConfig()) {
+    // Use a dynamic require here rather than a global require because this
+    // import path does not exist in eslint v7 which this library still
+    // supports
+    //
+    // Eslint exposes the new FlatESLint API under `eslint/use-at-your-own-risk` by
```
SimenB, line 122:
```suggestion
    // ESLint exposes the new FlatESLint API under `eslint/use-at-your-own-risk` by
```

SimenB (approved):

cprussin:
addressed your comments, thanks @SimenB

cprussin:
FYI that I just added a change to cache the ESLintConstructor value as I noticed the performance with this PR was quite poor due to traversing the filesystem to find the `eslint.config.js`  file on each lint run prior to adding the caching.

src/runner/runESLint.js
```diff
@@ -1,6 +1,13 @@
 const { ESLint } = require('eslint');
+const { version } = require('eslint/package.json');
+const findUp = require('find-up');
+const semver = require('semver');
+
 const getESLintOptions = require('../utils/getESLintOptions');
 
+const ESLINT_SUPPORTS_FLAT_CONFIG = semver.satisfies(version, '>=8');
+const ESLINT_FLAT_CONFIG_DEFAULT_FILENAME = 'eslint.config.js';
```
ljharb, line 9: i don't think this is needed since it's only used once, and is hardcoded in eslint?

cprussin: Yeah, I don't care too much, my thought process here was just to match the eslint implementation as closely as possible and they have it [extracted out to a const](https://github.com/eslint/eslint/blob/d85efad655deacc0dc3fdbbace33307094c3b91b/lib/eslint/flat-eslint.js#L92) as well but this is not something I have any opinion on and I'm happy to change it however if you care about it.

src/runner/runESLint.js
```diff
@@ -96,8 +103,70 @@ const getComputedFixValue = ({ fix, quiet, fixDryRun }) => {
...
+      case 'true':
+        return true;
+      case 'false':
+        return false;
+      default: {
+        const eslintFlatConfig = await findUp(
+          ESLINT_FLAT_CONFIG_DEFAULT_FILENAME,
+          { cwd: process.cwd() },
+        );
+        return !!eslintFlatConfig;
```
SimenB, line 118: Could we do `!= null` or something instead of double bang?

cprussin: Sure, done.

FWIW I wrote this function the way it is solely because it matches the [implementation](https://github.com/eslint/eslint/blob/d85efad655deacc0dc3fdbbace33307094c3b91b/lib/cli.js#L283-L301) in [eslint](https://github.com/eslint/eslint/blob/d85efad655deacc0dc3fdbbace33307094c3b91b/lib/eslint/flat-eslint.js#L266-L271) as closely as possible.  I'm happy to continue making changes here, but keep in mind that any  more changes will add cognitive overhead if they make changes there.  Also keep in mind that this is probably not the right solution anyhow, and that [hopefully eslint will expose some functionality that does this for us](https://github.com/eslint/eslint/issues/17129).

In any case, I'm happy to make any other changes you'd like, but keep in mind I didn't do it the way I did because I thought it was pretty / ideal, I did it to minimize differences with the reference implementation, and you'll need to decide if you care about that.

cprussin:
@ljharb @SimenB friendly nudge here -- anything else ya'll wanted to see here that I can polish off?

src/runner/runESLint.js
SimenB: fair point. thoughts on extracting all the "copied" code to its own file, including a comment pointing to where in ESLint's code it's from. Just to make future references easier

cprussin: @SimenB I think it's a great idea, however given that https://github.com/eslint/eslint/issues/17129 has traction and I have a [PR open](https://github.com/eslint/eslint/pull/17169) to implement it, at this point I figure it's probably better to just wait for that to land and swap this commit to use the soon-to-be-exposed `shouldUseFlatConfig` instead.

For backwards compatibility I'm thinking it's probably better to change https://github.com/jest-community/jest-runner-eslint/pull/176/files#diff-f5651cdb0e581e985fe919453f1a58dfba958912e3c5e7ddfb4f31057483eb17R8 to something along these lines instead:

```js
let FlatESLint;
let shouldUseFlatConfig;

try {
    eslintExperimental = require('eslint/use-at-your-own-risk');
    FlatESLint = eslintExperimental.FlatESLint;
    shouldUseFlatConfig = eslintExperimental.shouldUseFlatConfig;
} catch {
}

const eslintSupportsFlatConfig = () => FlatESLint !== undefined && shouldUseFlatConfig !== undefined;
```

That way, we wouldn't have to copy the eslint code in (I don't think we lose much by not supporting flat eslint config for intermediary versions between 8.0 and whenever https://github.com/eslint/eslint/pull/17169 gets released) and we don't have to encode version information to check for eslint compatibility, which feels cleaner to me.

Thoughts?  Cool with that plan?

ljharb: It occurs to me, though, that it's a breaking change to require eslint 8.whatever since we currently support all the current eslints - so it might be a good idea to keep this fallback implementation so that this PR can be semver-minor.

although i suppose just "not supporting flat config" in eslint 8's below 8.whatever isn't a breaking change, it's just a constraint on the new feature's usage.

cprussin: Yeah I mean, that's what I was saying in my comment, I don't think we lose all that much in not supporting the flat config (which is itself considered experimental anyways) for eslint 8.0.0 - whatever version releases my PR, and I really don't like being in the business of duplicating functionality.

The logic has also [changed at least once](https://github.com/eslint/eslint/pull/16356) through those versions (and maybe more, but I didn't look) so if we actually wanted to support all the patches/minors in between accurately, we'd need to compile those differences and account for them (although admittedly in this case I don't think it's a big deal to support the `ESLINT_USE_FLAT_COFIG` env var even though the eslint version doesn't support it, but still, I'm not sure if there were more significant changes).

I also don't think this is considered a breaking change since this functionality doesn't even currently exist in `jest-runner-eslint`, and using the implementation above it would gracefully degrade to it's exact current state on 8.x versions in between.  So it's really more of an optional new feature with a required minimum peer dependency than it is a breaking change, and there would be no reason to require a particular eslint version _unless_ someone wanted this functionality (in which case they should _already_ be using latest eslint anyways since this is pre-release experimental functionality in eslint).

SimenB: Agreed. Updating the OP in this PR to state the version requirement should be enough.

Thanks for pushing this through @cprussin!

cprussin:
@ljharb @SimenB [eslint v8.41.0](https://eslint.org/blog/2023/05/eslint-v8.41.0-released/) has been released, including my change to expose `shouldUseFlatConfig`.  I've rebased this PR on top of `main`, reverted changes to `package.json` and `yarn.lock`, and made [the change we discussed](https://github.com/jest-community/jest-runner-eslint/pull/176#discussion_r1187901587).  I've also upgraded eslint in the `yarn.lock` to v8.41.0 so that we're actually testing on the flat config code in CI.

I think the only deficiency in this PR at this point is that I didn't add any CI to test with eslint versions between v8.0.0 - v8.40.0.  I don't think that's necessary and it felt like overkill to me, but I'd be happy to go back and add it if ya'll feel it's important.

Other than that, I think it's good to ship.  Thanks again for the patience and help folks.

SimenB:
A single test for 8.40 could be nice, but if it proves hard I don't think it's a blocker 🙂 At least as long as we still test v7 I think it's highly unlikely we'll break anything

cprussin:
> A test run for 8.40 could be nice, but if it proves hard I don't think it's a blocker slightly_smiling_face At least as long as we still test v7 I think it's highly unlikely we'll break anything

No problem, added 8.40 to the test matrix.  It might prove to be overkill and wasted github action time but it's easy to strip out from the matrix config down the line.

cprussin:
@SimenB @ljharb either of you cool with merging this?  I'd love to get it upstreamed and be off my fork!

cprussin:
No worries at all, thanks for all the help getting this over the finish line!

SimenB:
https://github.com/jest-community/jest-runner-eslint/releases/tag/v2.1.0

Issue #58, opened 2018-10-23
cliOptions.fix is ignored


Radivarig:
```js
// jest-runner-eslint.config.js
module.exports = {
  cliOptions: {
    fix: true,
    rules: {
      'strict': 0,
    },
  },
}
```
The config is taken into account since strict override works, but it still does not fix and shows the following:
```
2 errors and 0 warnings potentially fixable with the `--fix` option.
```

PR #200, opened 2023-08-15
Fix issue #58 cliOptions.fix is ignored

hassoubeat:
Hi !

I have fixed [issue 58](https://github.com/jest-community/jest-runner-eslint/issues/58).

I just got it to work for now, so if you have any comments, please comment!

### DIFF
--- a/src/runner/__tests__/runESLint.test.js
+++ b/src/runner/__tests__/runESLint.test.js
@@ -1,7 +1,11 @@
 /* eslint-disable class-methods-use-this, global-require */
 const path = require('path');
 
-const runESLintRunnerWithMockedEngine = ({ mockOptions, runESLintOptions }) => {
+const runESLintRunnerWithMockedEngine = ({
+  mockOptions,
+  runESLintOptions,
+  extraOptions,
+}) => {
   jest.resetModules();
   jest.doMock('eslint', () => ({
     ESLint: class {
@@ -18,11 +22,16 @@ const runESLintRunnerWithMockedEngine = ({ mockOptions, runESLintOptions }) => {
       loadFormatter() {
         return Promise.resolve({ format() {} });
       }
+
+      // eslint-disable-next-line no-unused-vars
+      static outputFixes(report) {
+        return Promise.resolve();
+      }
     },
   }));
   const runESLint = require('../runESLint');
 
-  return runESLint({ extraOptions: {}, ...runESLintOptions });
+  return runESLint({ extraOptions: extraOptions || {}, ...runESLintOptions });
 };
 
 it('Requires the config setupTestFrameworkScriptFile when specified', async () => {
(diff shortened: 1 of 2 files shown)


src/runner/runESLint.js
```diff
@@ -122,6 +122,15 @@ const getComputedFixValue = ({ fix, quiet, fixDryRun }) => {
   return undefined;
 };
 
+const removeUndefinedFromObject = object => {
+  return Object.keys(object).forEach(v => {
+    if (object[v] === undefined) {
+      // eslint-disable-next-line no-param-reassign
```
hassoubeat, line 128: The code using Object.assign timed out the test.
Please comment if you have any other good code.

src/runner/runESLint.js
```diff
@@ -122,6 +122,15 @@ const getComputedFixValue = ({ fix, quiet, fixDryRun }) => {
...
+      delete object[v];
+    }
+  });
+};
```
ljharb, line 132: mutating `extraOptions` doesn't seem like a good idea
```suggestion
function removeUndefinedFromObject(object) {
  return Object.fromEntries(Object.entries(object).filter(([, value]) => typeof value !== 'undefined'));
};
```
and if the current supported node versions don't support Object.fromEntries and/or Object.entries, please use https://npmjs.com/object.fromentries and https://npmjs.com/object.entries.

ljharb (changes requested):
This would need tests to cover whatever's being fixed.

src/runner/runESLint.js
hassoubeat: Thanks for the comment!
I have incorporated your suggestion!

ljharb:
Looks like you need to run prettier.

hassoubeat:
@ljharb 

> This would need tests to cover whatever's being fixed.

Thanks for the comment.
I'd be glad to get your comments if I'm wrong in my plan on how to write test cases.

I am thinking that I could pass the [cliOptions](https://github.com/hassoubeat/jest-runner-eslint/blob/ad28de028a1bacb3621dc4af98bc1c6c0960ca5f/src/runner/runESLint.js#L168-L172) to the [mkTestResults](https://github.com/hassoubeat/jest-runner-eslint/blob/ad28de028a1bacb3621dc4af98bc1c6c0960ca5f/src/runner/runESLint.js#L51-L59) and make sure that the `cliOptions.fix` is not overwritten with undefined on the test code.

ljharb:
That seems like it would work.

src/runner/__tests__/runESLint.test.js
```diff
@@ -18,11 +22,16 @@ const runESLintRunnerWithMockedEngine = ({ mockOptions, runESLintOptions }) => {
       loadFormatter() {
         return Promise.resolve({ format() {} });
       }
+
+      // eslint-disable-next-line no-unused-vars
+      static outputFixes(report) {
+        return Promise.resolve();
+      }
```
ljharb, line 29:
```suggestion
      async static outputFixes(report) {
      }
```

ljharb (approved):
LGTM assuming the test fails prior to the change, and passes after

SimenB (approved):
👍

PR #205, opened 2023-09-02
Fix reporting test durations

### LINKED ISSUE #204: Files linted in random sequence
Jest normally runs test files in order of how long they took to run last time, with the longest-running first.

`jest-runner-eslint` doesn't seem to do this. Files are linted in a seemingly random order.

It appears that the runner is not outputting timing information to Jest, or for some reason Jest is not recording it. In Jest's cache folder, for a normal test run, there's a `perf-cache` file which records timings. e.g.:

```json
{
  "/modules/foo/test/methods.test.js":[1,931],
  "/modules/foo/test/other.test.js":[1,2955]
}
```

The last number represents the number of milliseconds the test file took to run.

In comparison, the `perf-cache` file created after an `jest-runner-eslint` run:

```json
{
  "/modules/foo/test/methods.test.js":[1,0],
  "/modules/foo/test/other.test.js":[1,0]
}
```

Every file has a recorded duration of 0, so Jest cannot sequence the files.

I'd be happy to make a PR if someone can point me in the right direction.

overlookmotel:
Fixes #204.

Jest appears to have changed the format of `TestResult`. What was `perfStats.duration` is now `perfStats.runtime`:

https://github.com/jestjs/jest/blob/edeecfae240bfd54b33f3c9eae156bd157447e5a/packages/jest-test-result/src/types.ts#L90-L122

Adding `runtime` property solves the problem.

I've left the `duration` property in place too in case that provides compatibility with older versions of Jest.

### DIFF
--- a/src/runner/runESLint.js
+++ b/src/runner/runESLint.js
@@ -73,6 +73,7 @@ const mkTestResults = ({
       start: startTime,
       end: endTime,
       duration: endTime - startTime,
+      runtime: endTime - startTime,
       slow: false,
     },
     skipped: numPassingTests === 0 && numFailingTests === 0,


SimenB:
Seems weird to me Jest has both of these - when would they differ?

src/runner/runESLint.js
```diff
@@ -73,6 +73,7 @@ const mkTestResults = ({
       start: startTime,
       end: endTime,
       duration: endTime - startTime,
+      runtime: endTime - startTime,
```
SimenB, line 76: both of these seems silly to me - I'd assume Jest could figure it out on its own... I'll dig into this a bit more tomorrow and attempt to figure out why this would be needed

SimenB: Fix in Jest is https://github.com/jestjs/jest/pull/14473, but that would only fix it for Jest 29 - this regressed in Jest 26

SimenB (approved):
thanks!

SimenB:
https://github.com/jest-community/jest-runner-eslint/releases/tag/v2.1.1

overlookmotel:
Thanks for tracking it down and getting this merged so swiftly.

Issue #211, opened 2023-09-18
`cliOptions: { quiet: true }`unexpectedly reports warnings


ken-kenware:
I have a project where we use `{ cliOptions: { quiet: true }}` in order to not report warnings in CI.

We noticed recently that if a file had an error in it, CI would report all of the errors/warnings as errors (similar to this discussion: https://github.com/jest-community/jest-runner-eslint/issues/21). However, because I have `quiet` set to true, I would expect that ESLint should only report the errors.

I did some digging and it looks like the issue is happening because on this line, ESLint is properly filtering the events as expected:
https://github.com/jest-community/jest-runner-eslint/blob/main/src/runner/runESLint.js#L238

But then we pass the unfiltered report here: https://github.com/jest-community/jest-runner-eslint/blob/main/src/runner/runESLint.js#L249

I'm happy to submit a PR to fix!

SimenB:
PR very much welcome 🙂

nsunga:
@ken-kenware hello :wave:

is it possible for you to break down your file directory structure for how this runner works?

Im not sure how `cliOptions` is injected into the `jest.config.js` thats passed into `jest` -- I might be misunderstanding the current README for the setup

PR #212, opened 2023-09-18
Do not report warnings as errors in quiet mode

ken-kenware:
Hello!

This PR addresses issue #211 

When the `quiet` cliOption is passed the call to ESLint's `getErrorResults` filters out warnings, which is expected. Prior to this PR we were using the full list of errors + warnings from ESLint to report the assertions, which incorrectly logged out warnings as error. This PR uses the filtered list instead if `cliOptions.quiet` is specified.

I tested this locally to make sure that it's working as expected, and also verified that the integration test also does not show the warnings.

### DIFF
--- a/integrationTests/__fixtures__/quiet-mode/.eslintrc.json
+++ b/integrationTests/__fixtures__/quiet-mode/.eslintrc.json
@@ -1,0 +1,5 @@
+{
+  "rules": {
+    "no-console": "warn"
+  }
+}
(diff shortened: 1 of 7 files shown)


ljharb (approved):
very subtle change, lgtm if it's good with @SimenB

SimenB (approved):
👍

SimenB:
https://github.com/jest-community/jest-runner-eslint/releases/tag/v2.1.2

PR #228, opened 2024-02-27
fix: implement fixes to work with eslint 8.57.0

cprussin:
This PR implements updates the eslint dependency to 8.57.0 and fixes the error `'reportUnusedDisableDirectives' has been removed. Please use the 'overrideConfig.linterOptions.reportUnusedDisableDirectives' option instead.`.

In doing this, I also made the following improvements:

1. use the new [`loadESLint`](https://eslint.org/docs/latest/integrate/nodejs-api#loadeslint) function to load `ESLint` or `FlatESLint`, rather than using our own logic
2. refactors the `normalizeConfig.js` file to have explicit separate config shapes for the legacy config and the flat config.  This also removes the need to have the code to deal with invalid config variables that we had [here](https://github.com/jest-community/jest-runner-eslint/blob/25a7d38b81c6618230f1f4c5153d639ab8aede08/src/runner/runESLint.js#L149-L162)

Note that #1 makes this library incompatible with using the flat config format with eslint < 8.57.0.  But I think that's OK given that 1) the flat config was experimental anyways and 2) if someone had a use case for locking eslint < 8.57.0, they could just as easily also lock jest-runner-eslint to v2.1.2.  It's technically a breaking change, but I don't think it absolutely must be marked as such.

### DIFF
--- a/src/runner/runESLint.js
+++ b/src/runner/runESLint.js
@@ -1,31 +1,5 @@
-const { ESLint } = require('eslint');
+const { ESLint, loadESLint } = require('eslint');
 const getESLintOptions = require('../utils/getESLintOptions');
-
-let FlatESLint;
-let shouldUseFlatConfig;
-
-try {
-  // Use a dynamic require here rather than a global require because this
-  // import path does not exist in eslint v7 which this library still
-  // supports
-  //
-  // ESlint exposes the new FlatESLint API under `eslint/use-at-your-own-risk` by
-  // using it's [export configuration](https://tinyurl.com/2s45zh9b).  However,
-  // the `import/no-unresolved` rule is [not aware of
-  // `exports`](https://tinyurl.com/469djpx3) and causes a false error here.  So,
-  // let's ignore that rule for this import.
-  //
-  // eslint-disable-next-line global-require, import/no-unresolved
-  const eslintExperimental = require('eslint/use-at-your-own-risk');
-  FlatESLint = eslintExperimental.FlatESLint;
-  shouldUseFlatConfig = eslintExperimental.shouldUseFlatConfig;
-} catch {
-  /* no-op */
-}
-
-if (shouldUseFlatConfig === undefined) {
-  shouldUseFlatConfig = () => Promise.resolve(false);
-}
 
 /*
  * This function exists because there are issues with the `pass`, `skip`, and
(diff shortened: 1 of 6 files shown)


cprussin:
ping @SimenB can I get your help merging & releasing this please?  It's a hard blocker for using eslint 8.57.0 with this lib so I'd love to get it out as soon as possible.  Thanks in advance!

src/runner/runESLint.js
```diff
@@ -1,32 +1,6 @@
-const { ESLint } = require('eslint');
+const { ESLint, loadESLint } = require('eslint');
 const getESLintOptions = require('../utils/getESLintOptions');
 
-let FlatESLint;
```
ljharb, line 4: why is all this important compat code being removed?

cprussin: As I'd mentioned in the PR description, this code is all superseded by `loadESLint` which was added in eslint 8.57.0 (and is exactly the API I'd originally requested to implement in https://github.com/eslint/eslint/issues/17129).  This code was only ever here because we weren't able to get the API added in eslint at the time.

The effect of removing the code is that the flat config won't work with versions of jest-runner-eslint after this change if eslint is pinned to a version between 8.41.0 and 8.56.0.

Even if we kept the compat code, we'd want to use the new `loadESLint` since it's the officially supported mechanism for doing this.  What we could do us use `loadESLint` if it exists, and fall back to the compat code if it doesn't.  However, the use case for moving eslint to 8.57.0, but locking jest-runner-eslint to a lower version, seems to me to make very little sense -- enough that I doubt it's worth maintaining the added complexity.  Given the flat config has been experimental anyways and was not guaranteed to be stable, I don't even think this should actually be considered a breaking change, although it _technically_ is.

We can add back the compat code if you felt strongly, but I seriously doubt there's any practical benefit of having it.

ljharb: That makes it a breaking change, so we'd want to keep the compat code regardless, and that's fine to use loadESLint and fall back.

Avoiding breaking changes has the largest possible practical benefit, higher than any features or bug fixes.

cprussin: Fair enough -- although it's worth noting that internals in ESLint [have been renamed](https://github.com/eslint/eslint/pull/17914) over those versions, were never guaranteed to be stable, and likely the compat code doesn't even work for the whole range.  We'd need to test for every version in between the range to know where it works and where it doesn't.  So even having the compat code doesn't guarantee we support versions 8.41.0+ -- we likely support 8.41.0 until some version where something changed in ESLint, and then as of this commit, we'd support 8.57.0+.

I'll happily re-add it to get this moving forward, but it won't add any practical value.

ljharb: We definitely should be testing for every minor version in between the range, regardless :-)

the practical value is that, even if untested, there are users who won't be broken by removing it.

cprussin: I suppose -- I mean, again, these were never guaranteed to be stable in the first place and have been actively breaking inside of ESLint itself in minor versions.  And the use case for moving this library without moving eslint is a bit absurd.  Again, I'm happy to add it back just to get this moving forward, but it feels very pedantic.

cprussin: Re added the compat code (although I added a slightly changed version that better matches the new 8.57.0 APIs, so that we can keep the rest of the code simpler and unaware of the backwards compat).

SimenB (approved):
thanks!

SimenB:
https://github.com/jest-community/jest-runner-eslint/releases/tag/v2.2.0

PR #291, opened 2025-06-27
feat: support Jest v30

G-Rath:


G-Rath:
ok so the E2E tests for this are failing:

image

Which I think is because of a bug in Jest / the new resolver introduced in v30 - the e2e tests are setting `../../../` as the runner, which Jest v29 apparently finds as a "node module" and its resolved to `/projects/jest-runner-eslint/build/runner/index.js`:

image

([that's this branch](https://npmfs.com/package/jest-resolve/29.7.0/build/utils.js#L126))

whereas in Jest v30 that doesn't happen resulting in `jest-config/build/index.js` being resolved for some reason, as the runner for some reason and then the test suite crashes as that is not a runner:

image

(the above logs are coming from a hacky file-based logger I put together to try and trace this which starts from the config normalizer [here](https://npmfs.com/package/jest-config/29.7.0/build/normalize.js#L684) - its literally just writing stuff to file in a `console.log` manner, with a basic attempt to do some tracking of the call stack - the patched I log calls are the same in my v30 and v29 clones right up until the "base resolver" part as that's where the two major versions actually differ).

@SimenB I think I've reached the point where I'll need to hand over to someone else (assumingly @JounQin) - I'm happy to create a new bug report over in Jest if folks are ok with this being the reproduction; I have already had a look though the issues, prs, and discussions, and couldn't find anything that looked like this is already known - I suspect its a bit of an edge case.

fwiw the latest version of this package already seems to work fine with Jest v30 as we're using it in eslint-plugin-jest

JounQin:
Where is `jest-runner-../../../` generated? `jest-resolve` or `jest-runner-eslint`?

G-Rath:
It comes from within Jest, just a couple of lines up from the return I linked: https://npmfs.com/package/jest-resolve/29.7.0/build/utils.js#L124

JounQin:
Then I think it's a reasonable bug to be fixed, please help to raise a new issue?

G-Rath:
@JounQin cool, I can probably find time to open an issue on the jest repo tomorrow, but also happy if you want to make one so you can work on it

JounQin:
Then it depends on should we consider `../../..` as a valid `runner` name. 🤔

cc @SimenB @cpojer

But why is that set like this at the first place? It seems just depending on unintended behaviour.

G-Rath:
To be clear, what's happening here is the runner is pointed to as a package i.e. ./../../ leads to the root of this repo which has a package.json so then jest/node looks at its main field and resolved the runner to it's actually js entry point.

That sounds reasonable to me, though there might be preference to have this done another way (I've not tried, but I expect using `npm link` or pointing at the actual js file (`../../../build/runner/index.js`) would also work).

However either way it feels like there's a separate bug with the fact we end up having a path leading to `jest-config` returned - there's nothing pointing to that afaik, so it feels like something has tried to do something akin to a `find-pkg-up` and the `jest-config` package is the first thing it finds that matches

JounQin:
Sorry I think I still can't fully understand the root cause, a failing test case to `jest-resolve` would be appreciated if you think it's a upstream issue.

JounQin:
@G-Rath I think you you need post a new issue on jest repository to get @SimenB and @cpojer involved.

G-Rath:
@JounQin yup I'll be following up with this after the weekend 🙂

G-Rath:
@SimenB this should be ready to go - I've got a completely green CI on my local and in my fork repo. There is a bit of a weird situation though regarding `create-jest-runner` as the latest 0.x version of that doesn't allow Jest v30 as a peer dependency even though it does seem to work fine with it (we've already been using Jest v30 on `eslint-plugin-jest` for bit), but you've done v1 with that change - so this PR will mean the peer constraints for `jest-runner-eslint` will be happy but not `create-jest-runner`...

Let me know what you'd like to do about that - I'll also look to open an issue on the Jest repo about the `../../../` change after we've gotten the ball rolling on this PR

ljharb (commented):
does this mean we're not testing an unresolved string for jest < 30?

G-Rath:
[the top level config](https://github.com/jest-community/jest-runner-eslint/blob/main/jest.config.js#L28) references the runner itself via an unresolved string and is working fine so there is still a sort of test covering that

ljharb:
how come the others need to be resolved for jest 30, but that one doesn't?

G-Rath:
it looks like it's because it is prefixed with `./` - happy to do that instead of use `require.resolve` if it'd make you happier

ljharb:
I'm not concerned with the usage of require.resolve; I'm concerned with the change of existing tests, which constitutes a breaking change. Is there a way we can make the path be conditional, and be resolved only in jest 30+?

G-Rath:
Sure we should be able to, but I don't think the concern is founded because:
1. the previous versions of Jest are not going to change
2. the expected way to use this package is with `jest-runner-eslint`, not to be providing a relative path to the package
3. this seems like a bug in Jest v30 so ideally it'll be a non-issue in the near future
4. it's likely a new major version will be needed anyway for bumping `create-jest-runner`

Confirming what changes are to be done for that last point is what I'm more interested in at this stage - I can push the conditional path stuff up tomorrow morning

ljharb:
If it's a bug in jest 30, can we wait til it's fixed before supporting it?

G-Rath:
I'd really prefer if we didn't since currently its very hard to use this package with NPM due to peer dependency enforcement (which I believe is made more complicated by the `create-jest-runner` dependency) - currently the smoke test workflow for `eslint-plugin-jest` is broken because we have to use NPM for installing for, and thus we run into this.

G-Rath:
If you'd like to try speed up getting it resolved in Jest I've still not had the time to open an issue officially reporting it there, so if you want to do that it'd really help

G-Rath:
I've opened https://github.com/jestjs/jest/issues/15712 with a reproduction - interestingly, it seems that the issue only happens when nesting at least three directories deep and when using Yarn

G-Rath:
@SimenB @ljharb there we go, green CI with nothing but version constraint changes 😄

ljharb (approved):
LGTM iff the answer to my comment is "yes"; otherwise this needs the version threshold for v30 bumped slightly.

SimenB (approved):
thank you! fun side quest with resolution 😀

PR #293, opened 2025-07-01
chore: add node 22 and 24 to CI matrix

SimenB:


.github/workflows/nodejs.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [12.x, 14.x, 16.x, 18.x, 19.x, 20.x]
+        node-version: [12.x, 14.x, 16.x, 18.x, 19.x, 20.x, 22.x, 24.x]
```
ljharb, line 21: we should look into using ljharb/actions/matrix here, so this can be automatic :-p

ljharb (approved):

